PR #20, opened 2022-05-31
Change default distfeed

gsanchietti:
Use distfeed pointing to [https://nextsec-testing.fra1.digitaloceanspaces.com/](https://nextsec-testing.fra1.digitaloceanspaces.com/index.html).

You can then execute:
```
opkg update
...
Downloading https://nextsec-testing.fra1.digitaloceanspaces.com/targets/x86/64/packages/Packages.gz
Updated list of available packages in /var/opkg-lists/nextsecurity_core
Downloading https://nextsec-testing.fra1.digitaloceanspaces.com/targets/x86/64/packages/Packages.sig
Signature check failed.
Remove wrong Signature file.
...
```

Please note that signature is not good.

### DIFF
--- a/config/distfeed.conf
+++ b/config/distfeed.conf
@@ -1,0 +1,1 @@
+CONFIG_VERSION_REPO="https://nextsec-testing.fra1.digitaloceanspaces.com"


gsanchietti:
During development, using original feed is much more convenience.
I'm going to postpone this one.

PR #24, opened 2022-06-21
Threat shield

gsanchietti:
Replace threat shield with:
- a NFT based solution for IP block lists
- a adblock based for DNS block lists

See also #22 

TODO:
- [x] support URL change (added documentation)
- [x] make sure to enable it after reboot
- [x] keep config file after removal
- [x] update NFT implementation to support multiple download URLs
- [x] threat shield DNS: try to add back the proxy bypass (see https://github.com/openwrt/packages/compare/master...NethServer:packages:adblock_bypass)


gsanchietti:
Ready for review @filippocarletti

gsanchietti:
See also nethesis/blacklists/pull/18

packages/ns-threat_shield/Makefile
```diff
@@ -0,0 +1,85 @@
...
+fi
+exit 0
+endef
+
+define Package/ns-threat_shield/prerm
+#!/bin/sh
+if [ -z "$${IPKG_INSTROOT}" ]; then
+    uci set threat_shield.config.status=0
+	uci commit
+	/usr/sbin/ts
```
filippocarletti, line 70: //usr/lib/opkg/info/ns-threat_shield.prerm: /usr/lib/opkg/info/ns-threat_shield.prerm-pkg: line 5: /usr/sbin/ts: not found
I guess it'd be `ts-ip`

filippocarletti (changes requested):
Fix package removal

filippocarletti (approved):

gsanchietti:
Bypass has been implemented as local patch to adblock.
See also #29

PR #42, opened 2022-10-11
Add support for Italian keyboard on console

filippocarletti:
* Load the keymap: loadkmap </usr/share/keymaps/it.map.bin
* To build a binary keymap on CentOS (as root): loadkeys -b it.map.gz -u -q >it.map.bin


files/lib/preinit/95_keymap
```diff
@@ -0,0 +1,4 @@
+# Copyright 2022 Nethesis
+
+KMAPDIR="/usr/share/keymaps"
+[ -n "$lang" ] && [ -s $KMAPDIR/"$lang".map.bin ] && cat $KMAPDIR/"$lang".map.bin | loadkmap
```
gsanchietti, line 4: Where is defined the `$lang` variable and what is the default value?

files/lib/preinit/95_keymap
```diff
@@ -0,0 +1,4 @@
+# Copyright 2022 Nethesis
```
gsanchietti, line 1: If you want to add an header, please see https://github.com/NethServer/nextsecurity/blob/master/README.md#license-headers

gsanchietti (changes requested):

files/lib/preinit/95_keymap
gsanchietti: Reading the description, I guess the variable comes from grub and the default is empty. Am I right?

filippocarletti: Right

files/lib/preinit/95_keymap
filippocarletti: Do we need a license in a one-line file? :-)

filippocarletti:
We still have to work on sysupgrade, we can maintain grub options.
AFAIK, no, there's no way to view the keyboard layout.

gsanchietti (approved):
Let's go for a very basic implementation for now.

PR #52, opened 2022-11-02
Allow the installation directly from the USB stick

NethNick:
Added compressed ext4 image to itself (in a VFAT partition)
Included install-nx.sh script to final image

### DIFF
--- a/builder/make.sh
+++ b/builder/make.sh
@@ -27,6 +27,5 @@ EOF
 /home/build/openwrt/staging_dir/host/bin/mcopy -i "store.part" ${IMG} ::nextsecurity.img.gz
 OFFSET=$(echo "$(fdisk -l extended.img| grep img3)" | cut -c 15-22)
 dd if="store.part" of="extended.img" bs=512 seek="$OFFSET" conv=notrunc
-gzip "extended.img"
-cp extended.img.gz ${IMG}
+gzip "extended.img" -c >  ${IMG}
 rm "store.part"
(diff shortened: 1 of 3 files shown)


files/usr/sbin/install-nx.sh
```diff
@@ -0,0 +1,40 @@
...
+           N=1; 
+        else
+           N=$(grep "${T##*/}" /proc/partitions | wc -l)
+        fi
+        M=$(mount | grep $T| wc -l)
+        S=$(df -t vfat /boot | tail -n 1| cut -d " " -f 1| tr "1" "3")
+        if [ $N -eq 1 ] && [ $M -eq 0 ]; then
+           mkdir mymount
+           mount -t vfat $S mymount
+           IMG=$(find mymount -name nextsecurity\*-x86-64-generic-ext4-combined-efi.img.gz| tail -n 1);
```
gsanchietti, line 24: I'd remove the `x86_64` match so we can handle also other targets:
```suggestion
           IMG=$(find mymount -name nextsecurity\*-generic-ext4-combined-efi.img.gz| tail -n 1);
```

builder/make.sh
```diff
@@ -7,3 +7,26 @@
 
 # Start parallel build
 make -j $(nproc) V=sc world
+
+IMG=$(find /home/build/openwrt/bin -name nextsecurity\*-x86-64-generic-ext4-combined-efi.img.gz);
```
gsanchietti, line 11: Can we loop on all "nextsecurity\*-generic-ext4-combined-efi.img.gz" images?
For now we only have x86-64 target, but in the future...who knows!

PR #55, opened 2022-11-08
NextSecurity: Allow the installation directly from the USB stick

NethNick:
Added compressed ext4 image to itself (in a VFAT partition) 
Included install-nx.sh script to final image with support for multiple firmwares

### DIFF
--- a/builder/make.sh
+++ b/builder/make.sh
@@ -27,6 +27,5 @@ EOF
 /home/build/openwrt/staging_dir/host/bin/mcopy -i "store.part" ${IMG} ::nextsecurity.img.gz
 OFFSET=$(echo "$(fdisk -l extended.img| grep img3)" | cut -c 15-22)
 dd if="store.part" of="extended.img" bs=512 seek="$OFFSET" conv=notrunc
-gzip "extended.img"
-cp extended.img.gz ${IMG}
+gzip "extended.img" -c >  ${IMG}
 rm "store.part"
--- a/builder/make.sh
+++ b/builder/make.sh
@@ -8,7 +8,7 @@
 # Start parallel build
 make -j $(nproc) V=sc world
 
-IMG=$(find /home/build/openwrt/bin -name nextsecurity\*generic-ext4-combined-efi.img.gz);
+for IMG in $(find /home/build/openwrt/bin -name nextsecurity\*generic-ext4-combined-efi.img.gz); do
 gunzip -c ${IMG} > extended.img
 dd if=/dev/zero bs=1M count=257 >> extended.img
 sed -e 's/\s*\([\+0-9a-zA-Z]*\).*/\1/' << EOF | fdisk extended.img
@@ -29,3 +29,4 @@ OFFSET=$(echo "$(fdisk -l extended.img| grep img3)" | cut -c 15-22)
 dd if="store.part" of="extended.img" bs=512 seek="$OFFSET" conv=notrunc
 gzip "extended.img" -c >  ${IMG}
 rm "store.part"
+done
--- a/builder/make.sh
+++ b/builder/make.sh
@@ -4,14 +4,14 @@
 # Copyright (C) 2022 Nethesis S.r.l.
 # SPDX-License-Identifier: GPL-2.0-only
 #
-set -e
+
 # Start parallel build
 make -j $(nproc) V=sc world
 
 for IMG in $(find /home/build/openwrt/bin -name nextsecurity\*generic-ext4-combined-efi.img.gz); do
-gunzip -c ${IMG} > extended.img
-dd if=/dev/zero bs=1M count=257 >> extended.img
-sed -e 's/\s*\([\+0-9a-zA-Z]*\).*/\1/' << EOF | fdisk extended.img
+   gunzip -c ${IMG} > extended.img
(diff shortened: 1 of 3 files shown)


NethNick:
Installation directly from the USB stick with VFAT partition for firmware

builder/make.sh
```diff
@@ -7,3 +7,25 @@
 
 # Start parallel build
 make -j $(nproc) V=sc world
+
+IMG=$(find /home/build/openwrt/bin -name nextsecurity\*-x86-64-generic-ext4-combined-efi.img.gz);
```
gsanchietti, line 11: Can you please avoid to look for only for x84_64 images?
We could just simply add a loop for all `nextsecurity\*-generic-ext4-combined-efi.img.gz` images

files/usr/sbin/install-nx.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
```
gsanchietti, line 1: Remember to always add the license header https://nethserver.github.io/nextsecurity/license_headers/.
```suggestion
#!/bin/bash
#
# Copyright (C) 2022 Nethesis S.r.l.
# SPDX-License-Identifier: GPL-2.0-only
#
```

files/usr/sbin/install-nx.sh
```diff
@@ -0,0 +1,60 @@
...
+              zcat $IMG| dd of=$T conv=notrunc
+           fi
+           umount /tmp/firmware
+           rmdir /tmp/firmware
+        else
+           if [ $M -eq 0 ]; then
+              echo -e "Multiple partitions find on target device, check it or use -f to force overwrite"
+           else
+              echo -e "Target partition in use, umount it first"
+           fi
```
gsanchietti, line 56: These are errors, the script should exit with non-zero code.
```suggestion
           if [ $M -eq 0 ]; then
              echo -e "Multiple partitions find on target device, check it or use -f to force overwrite"
           else
              echo -e "Target partition in use, umount it first"
           fi
           exit 1
```

files/usr/sbin/install-nx.sh
```diff
@@ -0,0 +1,60 @@
...
+        fi
+else
+        echo -e "Target device not found"
```
gsanchietti, line 59: These are errors, the script should exit with non-zero code.
```suggestion
        echo -e "Target device not found"
        exit 1
```

files/usr/sbin/install-nx.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+if [ $# -eq 0 ]; then
+    echo -e "No arguments supplied, target device for installation needed\n$0 -t /dev/sdX [-s source]"    
+    exit 1
+fi
+while getopts "ft::s::" opt; do
+            case $opt in
+            (f) F=1 ;; #Force write
+            (s) S=${OPTARG} ;; #Source image
+            (t) T=${OPTARG} ;; #Target disk
```
gsanchietti, line 10: Please use variables with one chars only on limited scope, like inside loops.
Generally speaking variable names should explain the purpose.
Also, as convention, variables in upper case are usually constant.

So the above code should be something like:
```
            (f) force=1 ;; #Force write
            (s) source=${OPTARG} ;; #Source image
            (t) target=${OPTARG} ;; #Target disk
```

files/usr/sbin/install-nx.sh
```diff
@@ -0,0 +1,60 @@
...
+           if [ ! -f $IMG ]; then
+              echo "Firmware not found"
+              umount /tmp/firmware
+              rmdir /tmp/firmware
+              exit 1
+           else
+              zcat $IMG| dd of=$T conv=notrunc
+           fi
+           umount /tmp/firmware
+           rmdir /tmp/firmware
```
gsanchietti, line 50: You could avoid duplicated on line (43-44) by setting a trap at the top of the file:
```
trap "umount /tmp/firmware; rmdir /tmp/firmware" EXIT
```

gsanchietti (commented):
Also remember to add the documentation inside [quickstart](https://github.com/NethServer/nextsecurity/blob/master/docs/quickstart.md).

files/usr/sbin/install-nx.sh
```diff
@@ -0,0 +1,60 @@
...
+              read -r IMG
+              IMG=${B[$IMG]}
+           fi
+           if [ ! -f $IMG ]; then
+              echo "Firmware not found"
+              umount /tmp/firmware
+              rmdir /tmp/firmware
+              exit 1
+           else
+              zcat $IMG| dd of=$T conv=notrunc
```
gsanchietti, line 47: The proposed change should speed up the write and avoid error if the image contains garbage at the end:
```suggestion
              zcat $IMG| dd of=$T  bs=64K iflag=fullblock conv=notrunc
```

NethNick (commented):
Fullt agree with suggestions

builder/make.sh
```diff
@@ -7,3 +7,25 @@
```
gsanchietti, line 7: Make sure to not execute other commands on failure:
```suggestion
set -e
```

gsanchietti (approved):

PR #68, opened 2022-11-25
menu: hide install options on installed machine

gsanchietti:
If a machine has been installed with dd, like VMs, hide the install options.
The fix should work also with machine with extra external storage.

The install option can still appear when connecting an extra storage which is not already mounted. This should be a temporary situtation and culd be ignored.

### DIFF
--- a/files/usr/sbin/menu
+++ b/files/usr/sbin/menu
@@ -4,6 +4,15 @@
 # SPDX-License-Identifier: GPL-2.0-only
 #
 
+function list_free_disks()
+{
+    for device in $(find /dev/ -name 'sd[a-z]' -o -name 'nvme[0-9]' -o -name 'hd[a-z]' -o -name 'vd[a-z]' -o -name 'xvd[a-z]' -o -name 'mmcblk[0-9]'); do
+	if ! grep -q $device /proc/mounts; then
+            echo $device
+	fi
+    done
+}
+
 function load_kmap() {
     A=0;declare -a B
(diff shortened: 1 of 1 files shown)


NethNick (approved):

files/usr/sbin/menu
```diff
@@ -69,12 +78,12 @@ function wrong_command(){
...
 function menu(){
    my_disk=$(df -t vfat /boot | tail -n 1| cut -d " " -f 1| tr -d "1")
-   disk=$(find /dev/ -name sd[a-z] -o -name nvme[0-9] -o -name hd[a-z] -o -name vd[a-z] -o -name xvd[a-z] -o -name mmcblk[0-9] | grep -v $my_disk | tail -n 1)
+   disk=$(list_free_disks | tail -n 1)
    echo -ne "
 Quick options:
 $(color_green '1)') Load alternate keyboard maps"
    label=($(block info ${my_disk}3));
-   if [[ "${label[2]}" == "LABEL=\"store\"" ]]; then
+   if [[ "${label[2]}" == "LABEL=\"store\"" && "$disk" != "" ]]; then
```
NethNick, line 86:
```suggestion
   usb=$(lsusb -t)
   if [[ "${label[2]}" == "LABEL=\"store\"" ]] && [[ $usb = *"usb-storage"* ]]; then 
```

NethNick (changes requested):
Changed check from "empty disk" to "lsusb usb-storage device"

PR #71, opened 2022-11-25
conf: add debug tool whois

gsanchietti:
The command is usefull to check public IPs info especially when such IPs are blocked by filters


filippocarletti (commented):
Did you try busybox's whois command?
It'd be my preferred choice.
Note: whois is not installed by default in NethServer 7 and I never added it to my systems.

gsanchietti:
> Did you try busybox's whois command?

No, I didn't but even busysbox one is not present.

If we want to add it, why not using the original one rather than a limited version?

>  Note: whois is not installed by default in NethServer 7 and I never added it to my systems.

I just used it to check if DPI was working good. :)

I agree we can go without it. If anyone needs it, it's possible to install from official OpenWrt repos.

PR #88, opened 2022-12-22
ns-don: disable sshd

gsanchietti:
Make sure OpenSSH doesn't try to start at boot

### DIFF
--- a/files/etc/uci-defaults/99-nextsec-sshd
+++ b/files/etc/uci-defaults/99-nextsec-sshd
@@ -1,0 +1,2 @@
+service sshd stop || return 0
+service sshd disable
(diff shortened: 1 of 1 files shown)


gsanchietti:
Current patch doesn't work: the service is not disabled. 
Moving the command inside ns-don uci-default doesn't resolve the issue.

NethNick (commented):
Want something earlier than level 99 ?

gsanchietti:
> Want something earlier than level 99 ?

I do not think position is important: put it where it fit best.

> We could disable dropbear from running with:

We do not have to disable dropbear, but OpenSSH

gsanchietti:
Replaced by #96

PR #89, opened 2022-12-27
rsyslog: early start at 12 to improve boot logging

NethNick:
Changed startup level of rsyslog from 20 to 12, improving local (or eventually) remote logging


packages/rsyslog/files/12_rsyslog
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
```
gsanchietti, line 1: Why this file has been moved from 20 to 12?

NethNick: From testing and inspection this level (same as original logd) catches the most complete log and we decided this was enough.

gsanchietti: Change to this file is not related to the boot order. Please revert this change.

Change on init script looks good.

gsanchietti (approved):

PR #90, opened 2022-12-28
Firewall objects

gsanchietti:
The implementation is only available for DPI rules.

PoC of firewall objects.

See [README](https://github.com/NethServer/nextsecurity/blob/ns-objects/packages/ns-objects/README.md) for more info

TODO:
- [x] documentation
- [x] remove support for user/groups as destination, evaluate only records of rule type
- [x] abs2fw: rewrite `/etc/config/firewall` only when needed (see https://docs.python.org/3/library/filecmp.html)
- [x] implement user and groups inside netifyd 
- [x] ~~optimize map2ip using nftables bindings~~ the speedup is too low: around a tenth  of a second for adding 254 IPs
- [x] evaluate how to implement users/groups also inside dnsdist: it's possible by creating complex rules based on IPs or MAC addresses which can be forwarded to upstream server https://blog.powerdns.com/2016/01/27/per-device-dns-settings-selective-parental-control/


packages/ns-objects/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+A user is a dynamic entity representing all physical and devices belonging to a person like PCs, mobile phones or VPN road warrior accesses.
+A group is a dynamic entity representing a list of users.
+
+The `user` and `group` object are identified by the section name.
+The section name must:
+
+- be unique
+- be a valid UCI id (it may contain only the characters `a-z`, `0-9` and `_`)
+- have a maximum length of 14 characters (nft set name must be 16 characters or less, 4 chars are reserved by `abs2f2` to generate unique nft set names)
```
filippocarletti, line 24:
```suggestion
- have a maximum length of 14 characters (nft set name must be 16 characters or less, 4 chars are reserved by `abs2fw` to generate unique nft set names)
```

filippocarletti: Max len is 14 or 12?

packages/ns-objects/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+The `user` object can have the following non-mandatory options:
+
+- `name`: name of user
+- `description`: a longer label for the user
+- `macaddr`: list of MAC address that will be resolved as IPs (see below)
```
filippocarletti, line 30:
```suggestion
- `macaddr`: list of MAC addresses that will be resolved as IPs (see below)
```

packages/ns-objects/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+config group 'vip'
+	option description 'Very Important People'
+	list user 'goofy'
+	list user 'daisy'
+```
+
+### Abstract firewall
+
+The `absfirewall` configuration has the same syntax of `firewall` configuration but is also supports the following options inside the `rule` record:
```
filippocarletti, line 66:
```suggestion
The `absfirewall` configuration has the same syntax of `firewall` configuration but it also supports the following options inside the `rule` record:
```

packages/ns-objects/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+- `user_src`: match incoming traffic from the specified user IPs
+- `group_src`: match incoming traffic from the specified user group IPs
+
+The abstract firewall must be converted to a standard firewall configuration using the `abs2f2` command. The command will:
```
filippocarletti, line 71:
```suggestion
The abstract firewall must be converted to a standard firewall configuration using the `abs2fw` command. The command will:
```

packages/ns-objects/README.md
```diff
@@ -0,0 +1,109 @@
...
+```
+abs2fw > /var/etc/mac2ip.json
+service mac2ip reload
+fw4 restart
+```
+
+## Host
+
+Objects of type `host` are used only as labels inside the UI to display details on firewall rules source and destinations.
+This kind of objects can't be used inside `absfirewall` not `firewall` config: if the user changes the IP address of an host object, such
```
filippocarletti, line 101:
```suggestion
This kind of objects can't be used inside `absfirewall` or `firewall` config: if the user changes the IP address of an host object, such
```

packages/ns-objects/README.md
```diff
@@ -0,0 +1,72 @@
+# ns-objects
+
+This package manages the firewall objects.
+
+Supported object types:
+- `user`
+- `group`
+- `host`
+
+All objects are saved inside `/ect/config/objects` UCI database.
```
filippocarletti, line 10:
```suggestion
All objects are saved inside the `/etc/config/objects` UCI database.
```

packages/ns-objects/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+A user is a dynamic entity representing all physical and virtual devices belonging to a person like PCs, mobile phones or VPN road warrior accesses.
+A group is a dynamic entity representing a list of users.
+
+The `user` and `group` object are identified by the section name.
+The section name must:
+
+- be unique
+- be a valid UCI id (it may contain only the characters `a-z`, `0-9` and `_`)
+- have a maximum length of 12 characters (nft set name must be 16 characters or less, 4 chars are reserved for future use
```
filippocarletti, line 22:
```suggestion
- have a maximum length of 12 characters (nft set name must be 16 characters or less, 4 chars are reserved for future use)
```

packages/ns-objects/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+The `user` object can have the following non-mandatory options:
+
+- `name`: name of user
+- `description`: a longer label for the user
+- `macaddr`: list of MAC addresses that will be resolved as IPs (see below)
```
filippocarletti, line 28:
```suggestion
- `macaddr`: list of MAC addresses belonging to the user's devices
```

filippocarletti (changes requested):

filippocarletti (approved):

PR #93, opened 2023-01-11
Fix threat shield IPs bug

gsanchietti:
After latest refactor, nft rules were duplicated.

### DIFF
--- a/packages/ns-threat_shield/files/ts-ip-download
+++ b/packages/ns-threat_shield/files/ts-ip-download
@@ -23,6 +23,7 @@ categories=$(uci -q get threat_shield.config.categories)
 if [ -z "$categories" ]; then
     exit 0
 fi
+categories="$categories whitelist"
 
 
 def_sources="/usr/share/threat_shield/nethesis-ip.sources.gz"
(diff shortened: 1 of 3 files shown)


packages/ns-threat_shield/files/ts-ip-download
```diff
@@ -23,6 +23,7 @@ categories=$(uci -q get threat_shield.config.categories)
 if [ -z "$categories" ]; then
     exit 0
 fi
+categories="$categories whitelist"
```
filippocarletti, line 26: "whitelist" can be downloaded if the system has an active subscription
This line could be moved at 43

filippocarletti (changes requested):

packages/ns-threat_shield/files/ts-ip-download
gsanchietti: The whitelist could be present also in other sources not only on nethesis ones.
If the category doesn't exists the download will fail but the process will continue.

In the end, I do not think we should restrict whitelist implementation only for Nethesis subscriptions.

gsanchietti: Can I move to the merge?

filippocarletti: Yes, go ahead :-)

PR #96, opened 2023-01-17
Disable OpenSSH server

gsanchietti:
Make sure OpenSSH doesn't try to start at boot.

This solution is quite ugly but I didn't find a better way.

See also https://github.com/NethServer/nextsecurity/pull/88


filippocarletti:
I'd avoid forking openssh to silence an error message.

gsanchietti:
I agree, let's keep the cosmetic issue for now.

PR #98, opened 2023-01-20
Refactor installation procedure

gsanchietti:
- Remove extra partition: reduce the image size by a half, installation now requires Internet access
- Remove install option from menu: the procedure will be documented
- Remove 'Exit' option from menu since it's not needed
- Install script renamed from `install-nx.sh` to `ns-install`
- Install script is now interactive (code partially imported from menu script)
- Install script can be executed in batch mode: `ns-install -t /dev/sdb -f`
- Add `ns-reset` script to execute factory default, the script requires Internet access to download the image
- Add `ns-download` script to download current or latest image version 
- Remove login menu


files/usr/bin/ns-download
```diff
@@ -0,0 +1,83 @@
...
+    case $opt in
+        (f) force=1 ;; # force download
+        (l) latest=1 ;; # download latest image
+        (*) help ;;
+    esac
+done
+
+# Force download by deleting destination dir
+if [ "$force" -eq 1 ]; then
+    rm -rf "$DL_DIR" &> /dev/null
```
filippocarletti, line 43:
```suggestion
    rm -rf "$DL_DIR"
```

files/usr/bin/ns-download
```diff
@@ -0,0 +1,83 @@
...
+img_name="nextsecurity-$version-$ARCH-generic-ext4-combined-efi.img.gz"
+img="$DL_DIR/$img_name"
+img_url="$BASE_URL/$version/targets/$OPENWRT_BOARD/$img_name"
+hash_url="$BASE_URL/$version/targets/$OPENWRT_BOARD/sha256sums"
+
+# Download if image does not exists
+if [ ! -d "$DL_DIR" ] || [ ! -f "$img" ] || [ ! -f "$hash" ]; then
+    mkdir -p "$DL_DIR"
+    rm -f "$img" "$hash" &> /dev/null
+    curl -L -s -m 60 -o "$hash" "$hash_url"
```
filippocarletti, line 66: 60 seconds to download an hash seems like an eternity :-)

files/usr/bin/ns-download
```diff
@@ -0,0 +1,83 @@
...
+    curl -L -s --connect-timeout 60 -o "$img" "$img_url"
```
filippocarletti, line 67: Also 60 seconds for the connect is a really big time

files/usr/bin/ns-download
gsanchietti: What about 30?

files/usr/bin/ns-download
gsanchietti: What about 20?

files/usr/bin/ns-download
gsanchietti: fixed

files/usr/bin/ns-download
gsanchietti: fixed

PR #126, opened 2023-06-09
ns-report: add dpireport daemon

gsanchietti:
Lightsquid-like reports based on netifyd data streams.


gsanchietti:
With latest ndpi (4.4.3), the prototype does not work anymore.
It seems netifyd is not raising flow_purge events.

gsanchietti:
Updated for netifyd 4.4.3

filippocarletti (approved):

PR #141, opened 2023-07-24
Added config file for ns-api

Tbaile:
Added defaults for Multi Wan feature


packages/ns-api/Makefile
```diff
@@ -24,6 +24,10 @@ define Package/ns-api
 	DEPENDS:=+python3-nethsec +python3-openssl +python3-urllib +python3-idna
 	PKGARCH:=all
 endef
+
+define Package/ns-api/conffiles
+/etc/config/ns-api
```
gsanchietti, line 29: Indent with a tab here

packages/ns-api/files/config
```diff
@@ -0,0 +1,40 @@
...
+        # ping size and time to live, leave defaults if possible
+        option ping_size '56'
+        option ping_max_ttl '60'
+        # after how much time a ping should be considered failed (in seconds)
+        option ping_timeout '4'
+        # ping interval on a online interface (in seconds)
+        option ping_interval '10'
+        # if ping fails, this is how often we check for an online status (in seconds)
+        option ping_failure_interval '5'
+        # when an ip finally pings, how frequent a ping should be sent (in seconds)
```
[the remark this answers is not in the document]
Tbaile: Totally agree, is this better?
```suggestion
        # seconds between pings when recovering from a failure, below a better explanation on how this option is treated
```

packages/ns-api/Makefile
Tbaile: Thought to add the tab too, [found out that is not possible](https://openwrt.org/docs/guide-developer/packages#packageconffiles_optional)

gsanchietti: Whoa! We should fix other packages!

gsanchietti (approved):

PR #158, opened 2023-09-13
DHCP APIs

gsanchietti:
New APIs for `ns.dhcp`:
- list-interfaces
- edit-interface
- list-dhcp-options
- get-interface
- list-active-leases
- list-static-leases
- get-static-lease
- delete-static-lease
- add-static-lease
- edit-static-lease

### DIFF
--- a/packages/ns-api/README.md
+++ b/packages/ns-api/README.md
@@ -1106,6 +1106,260 @@ Error response:
 {"error": "no subscription info found"}
 ```
 
+## ns.dhcp
+
+Manage DHCPv4 servers and static leases.
+
+### list-interfaces
+
+List all interfaces where it is possible to enable a DHCPv4 server:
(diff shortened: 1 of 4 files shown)


packages/ns-api/README.md
```diff
@@ -1106,6 +1106,260 @@ Error response:
...
+    }
+  ],
+  "first": "192.168.100.2",
+  "last": "192.168.100.150",
+  "leasetime": "12h",
+  "active": true
+}
+```
+
+Eeach element of the `options` array is a key-value object.
```
edospadoni, line 1204:
```suggestion
Each element of the `options` array is a key-value object.
```

edospadoni (changes requested):

edospadoni (approved):

PR #173, opened 2023-09-26
Fix subscription API & doc

andre8244:
- Subscription info: don't throw an error if there is no subscription
- Subscription register: write UCI database only after a successful registration
- Fix documentation

### DIFF
--- a/packages/ns-plug/files/register
+++ b/packages/ns-plug/files/register
@@ -35,25 +35,23 @@ fi
 case "$type" in
     community)
         url="https://my.nethserver.com/api/"
-        ;;
-    enterprise)
-        url="https://my.nethesis.it/api/"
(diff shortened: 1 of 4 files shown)


packages/ns-plug/files/register
```diff
@@ -34,24 +34,16 @@ fi
 # Setup URLs
 case "$type" in
     community)
```
gsanchietti, line 36: There 3 checks of the type:
- one line 35
- one on line 47
- one on line 64

Please refactor the code to remove at least one check:
- we could merge the first and second one
- use only one operator type: if or switch

gsanchietti (changes requested):

gsanchietti (approved):

PR #179, opened 2023-09-29
Changes to ns.redirects API

andre8244:
- Return `log` and `reflect` in `list-redirects`


packages/ns-api/files/ns.redirects
```diff
@@ -56,6 +56,10 @@ def list_redirects():
         redirect["name"] = r.get("name", f"PF{count}") # name
         redirect["wan"] = r.get("src_dip", "any") # wan IP
         redirect["enabled"] = r.get("enabled", "1") == "1" # status
+        redirect["log"] = r.get("log", "1") == "1"
+        redirect["reflection"] = r.get("reflection", "1") == "1"
```
gsanchietti, line 60: Default for both options should be disabled:
```suggestion
        redirect["log"] = r.get("log", "0") == "1"
        redirect["reflection"] = r.get("reflection", "0") == "1"
```

gsanchietti (changes requested):

gsanchietti (approved):
If needed, rebase before merge

PR #183, opened 2023-10-06
ns-api: fix ns.dhcp & ns.dns

andre8244:
- ns.dhcp: API
- ns.dns: fix API
- fix documentation

### DIFF
--- a/packages/ns-api/files/ns.dns
+++ b/packages/ns-api/files/ns.dns
@@ -101,7 +101,8 @@ def get_record(args):
 def delete_record(args):
     u = EUci()
     try:
-        u.get('dhcp', args['record'])
+        record = u.get_all('dhcp', args['record'])
+        remove_wildcard(u, record.get("ip", ''), record.get('name', ''))
         u.delete('dhcp', args['record'])
         u.save('dhcp')
     except:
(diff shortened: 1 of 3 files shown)


packages/ns-api/files/ns.dns
```diff
@@ -52,22 +53,25 @@ def remove_wildcard(u, ip, name):
         cur = u.get('dhcp', server, 'address', list=True)
```
gsanchietti, line 53: It should be enough to revert other changes and just convert the tuple to list:
```suggestion
        cur = list(u.get('dhcp', get_server(u), 'address', list=True))
```

packages/ns-api/files/ns.dns
```diff
@@ -52,22 +53,25 @@ def remove_wildcard(u, ip, name):
...
-    u.set('dhcp', server, 'address', new)
+    u.set('dhcp', server, 'address', ",".join(new))
     u.save('dhcp')
 
 def add_wildcard(u, ip, name):
     remove_wildcard(u, ip, name)
     server = get_server(u)
+    address_list = []
     try:
-        addr = u.get('dhcp', server, 'address', list=True)
```
gsanchietti, line 66:
```
        addr = list(u.get('dhcp', server, 'address', list=True))
```

gsanchietti (approved):

PR #194, opened 2023-10-13
Add dpi list popular

Tbaile:
- python3-nethsec: bump version to 0.0.9
- ns-api: added popular filters
- ns-api: added list-popular call


packages/ns-api/files/ns.dpi
```diff
@@ -37,7 +37,11 @@ if cmd == 'list':
             'applications': 'str',
             'protocols': 'str'
         },
-        'list-devices': {}
+        'list-devices': {},
+        'list-popular': {
+            'limit': 32,
+            'page': 32
```
gsanchietti, line 43: Is this really needed?

Tbaile: Not really, I thought of removing it too, however since we're gonna change the UI sometimes in the future, we can still use the function without additional changes

gsanchietti: ok

gsanchietti (approved):

PR #227, opened 2023-11-14
Add nginx reverse proxy APIs

Tbaile:
Card: https://trello.com/c/6npknzJ7/180-reverse-proxy


packages/ns-api/files/ns.reverseproxy
```diff
@@ -0,0 +1,329 @@
...
+                if 'allow' in data:
+                    e_uci.set('nginx', location, 'allow', data['allow'])
+                # add location to _lan server
+                includes = list(e_uci.get('nginx', '_lan', 'include', list=True))
+                if 'conf.d/_lan.proxy' not in includes:
+                    includes.append('conf.d/_lan[.]proxy')
+                    e_uci.set('nginx', '_lan', 'include', includes)
+                # submit changes
+                __write_changes(e_uci)
+                print(json.dumps({'message': 'success'}))
```
gsanchietti, line 140: Why not returning the location id?
Maybe it can be more useful to the ui?

gsanchietti (commented):
Maybe the UI can also require a `get-proxy`?

packages/ns-api/files/ns.reverseproxy
```diff
@@ -15,38 +15,21 @@ from nethsec import utils
...
 def __create_location(uci, uci_server, location, proxy_pass):
     location_id = utils.get_random_id()
     uci.set('nginx', location_id, 'location')
     uci.set('nginx', location_id, 'uci_server', uci_server)
     uci.set('nginx', location_id, 'location', location)
-    uci.set('nginx', location_id, 'proxy_pass', proxy_pass)
+    prefix = proxy_pass.split('://')[0]
+    address = proxy_pass.split('://')[1]
+    # checking if address is a domain or an ip4/6 address
+    if address.count('.') == 3 or address.count(':') == 7:
```
gsanchietti, line 26: This could match also a host named "my.host.org".
I'd use something like:
```python
try:
    ipaddress.ip_address(address)
   e_uci.set('nginx', location_id, 'proxy_pass', proxy_pass)
except:
   e_uci.set('nginx', location_id, 'resolver', '127.0.0.1')
   e_uci.set('nginx', location_id, 'set', f'$upstream {address}')
   e_uci.set('nginx', location_id, 'proxy_pass', f'{prefix}://$upstream')
```

Otherwise we could always set the resolver, probably it should work also with IP addresses.

packages/nginx/files/nginx-proxy-gen
```diff
@@ -5,13 +5,18 @@
...
+
 import os
-import sys
 from glob import glob
+
 from euci import EUci
 
 u = EUci()
 
+
```
gsanchietti, line 19: Automatic prettier, I hate you! :D

Please avoid unnecessary changes: the patch it's already quite big

packages/nginx/README.md
```diff
@@ -41,7 +41,7 @@ Example of a path rule for the default virtual host:
 config location 'ns_location1'
 	option uci_server '_lan'
 	option proxy_pass 'https://192.168.100.234'
-	option description 'Reverse proxy with path'
+	option uci_description 'Reverse proxy with path'
```
gsanchietti, line 44: Is this really needed?
If the change is required, we should also change:
- https://github.com/NethServer/nethsecurity/blob/main/packages/ns-migration/files/scripts/reverse_proxy
- https://github.com/NethServer/nethserver-firewall-migration/blob/master/root/usr/share/nethserver-firewall-migration/reverse_proxy

packages/ns-api/files/ns.reverseproxy
```diff
@@ -0,0 +1,331 @@
...
+                if data['path'] in paths:
+                    raise ValidationError('path', 'unique', data['path'])
+                if 'destination' not in data:
+                    raise ValidationError('destination', 'required')
+                if not data['destination'].startswith('http://') and not data['destination'].startswith('https://'):
+                    raise ValidationError('destination', 'invalid', data['destination'])
+                if 'description' not in data or data['description'] == '':
+                    raise ValidationError('description', 'required')
+                if 'allow' in data and not isinstance(data['allow'], list):
+                    raise ValidationError('allow', 'invalid', data['allow'])
```
gsanchietti, line 111: Should we also add the validation to avoid the same path?

packages/ns-reverse-proxy/Makefile
```diff
@@ -1,41 +0,0 @@
-#
```
gsanchietti, line 1: If we are going to merge the packages, we should remove the old one from the build config: https://github.com/NethServer/nethsecurity/blob/main/config/ns-reverse-proxy.conf

gsanchietti (changes requested):

packages/nginx/README.md
Tbaile: Not required, but heavily suggested. The nginx config generator excludes any fields from the final configuration with `uci_` prefix, so it's just standardizing the behaviour

packages/ns-api/files/ns.reverseproxy
Tbaile: Already there! [Just a few lines above](https://github.com/NethServer/nethsecurity/pull/227/files#diff-1a04a0ae222beb786726104911ef0bc7a5097be32a40699b9bc70382400cbdaeR103) :smile:

packages/ns-reverse-proxy/Makefile
Tbaile: Was sure something like this was needed, what are the steps?

packages/nginx/files/nginx-proxy-gen
Tbaile: Whops, my bad :sweat:

packages/nginx/README.md
gsanchietti: So let's keep the change, but we must fix it also in both migration packages

Tbaile: K, will open a PR over there and update `ns-migration` here

packages/ns-api/files/ns.reverseproxy
gsanchietti: Sorry I didn't spot it. :+1:

packages/ns-reverse-proxy/Makefile
gsanchietti: Just remove the `ns-reverse-proxy.conf` file from the `config` dir.

gsanchietti (approved):

PR #236, opened 2023-11-27
config: add uvol and autopart

gsanchietti:
[ExPeRiMeNtAl]


gsanchietti:
Already tested by Filippo: It seems that uvol doesn't play well with ns-install.
Dropping the experiment.

PR #237, opened 2023-11-27
files: ns-install, call firstboot

gsanchietti:
Make sure the system is clean before installing it


filippocarletti (commented):
firstboot alters the running system, not the installed
Let's close this.

PR #265, opened 2024-01-05
QoS APIs

Tbaile:
Card: https://trello.com/c/P8EzabNk/174-qos-configuration

### DIFF
--- a/packages/ns-api/files/ns.qos
+++ b/packages/ns-api/files/ns.qos
@@ -1,0 +1,144 @@
+#!/usr/bin/python3
+
+#
+# Copyright (C) 2023 Nethesis S.r.l.
+# SPDX-License-Identifier: GPL-2.0-only
+#
+
+import json
+import numbers
+import sys
+
+from euci import EUci
+from nethsec import utils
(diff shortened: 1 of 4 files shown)


packages/ns-api/files/ns.qos
```diff
@@ -0,0 +1,144 @@
...
+        'delete': {
+            'interface': 'String'
+        },
+        'edit': {
+            'interface': 'String',
+            'disabled': False,
+            'upload': 0,
+            'download': 0,
+        },
+        'set-disabled': {
```
gsanchietti, line 54: What about changing the name of this function to something like `toggle-status`?

Tbaile: I've given thought about that, however `toggle-status` it's inferred that no additional data needs to be passed. With `set-status`, it should be hinted by the api call name.

Just a preference, pick you you like :man_shrugging:

gsanchietti: `set-status` sounds good!

gsanchietti (approved):

PR #294, opened 2024-01-26
Reverse Proxy migration revamp

Tbaile:
improved performance to list-proxies
fixed issue when listing proxies with pending certificates
moved all core functions of reverse proxy api to lib
using lib functions in reverse proxy migration to make everything omogeneous

- depends on: https://github.com/NethServer/python3-nethsec/pull/28

### DIFF
--- a/packages/ns-api/files/ns.reverseproxy
+++ b/packages/ns-api/files/ns.reverseproxy
@@ -559,7 +559,8 @@ elif cmd == 'call':
                         'destination': __parse_proxy_pass(e_uci, entry_id)
                     }
                     for certificate in certificates:
-                        if certificates[certificate]['cert_path'] == server['ssl_certificate']:
+                        if ('cert_path' in certificates[certificate]
+                                and certificates[certificate]['cert_path'] == server['ssl_certificate']):
                             item['certificate'] = certificate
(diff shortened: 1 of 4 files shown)


packages/ns-api/files/ns.reverseproxy
```diff
@@ -428,18 +294,11 @@ elif cmd == 'call':
                 raise ValidationError('description', 'required')
             if 'allow' in data and not isinstance(data['allow'], list):
                 raise ValidationError('allow', 'invalid', data['allow'])
+
+            # normalize optional parameters
+            allow = data['destination'] if 'allow' in data else []
```
gsanchietti, line 299: Why setting 'destination' as allow?
I think this one should be:
```
allow = data.get('allow', [])
```

Same on line 319.

gsanchietti (changes requested):

packages/ns-api/files/ns.reverseproxy
Tbaile: Definetly a typo, most fun part is that the migration is still successful (?)
image
 
:shrug:

gsanchietti (approved):

PR #311, opened 2024-02-05
ns-api: dedalo, handle vlan device

gsanchietti:
Address the following issues:
- when a hotspot relies on a VLAN, its registration fails to capture the MAC address in the Dedalo API's unit section.
- once a hotspot associated with a VLAN is unregistered, the VLAN remains stuck in the interface list and cannot be deleted.

Card: https://trello.com/c/hLhHR4F0/278-hotspot-vlan-management

### DIFF
--- a/packages/ns-api/files/ns.dedalo
+++ b/packages/ns-api/files/ns.dedalo
@@ -140,6 +140,8 @@ def list_parents():
         return {"parents": []}
 
 def unregister():
+    u = EUci()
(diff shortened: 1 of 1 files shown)


cotosso:
I tried the artifact image.
* registration still fails (missing MAC address)
* unregister works flawlessy, vlan interface is put down,  but still present in networks database as expected

gsanchietti:
Better than expected!

PR #331, opened 2024-02-12
ns-install: halt on completion

filippocarletti:
* halt instead of reboot after installation
* use lsblk to list available disks
* skip 0 sized disk (empty sd card readers)
* copy only the size of source img

### DIFF
--- a/files/usr/sbin/ns-install
+++ b/files/usr/sbin/ns-install
@@ -1,37 +1,13 @@
 #!/bin/bash
 #
-# Copyright (C) 2022 Nethesis S.r.l.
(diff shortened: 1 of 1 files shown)


files/usr/sbin/ns-install
```diff
@@ -79,47 +63,30 @@ done
 
 # select existing disks
 if [ -z "$target" ]; then
-    echo "Select target disk:"
+    echo "Select target disk (or any other key to exit):"
```
gsanchietti, line 66:
```suggestion
    echo "Select target disk (or press any other key to exit):"
```

gsanchietti (approved):

PR #333, opened 2024-02-14
Add notification system

gsanchietti:
Basic notification implementation.

See https://trello.com/c/XNOg8Ej8/111-dashboard-alerts-and-bell-notifications

TODO:
- [x] replace `priority` with `level`, valid levels are the syslog ones
- [ ] evaluate if the active flag is needed: maybe we could just delete read notifications
- [ ] a new mockup UI is needed to create a better backend implementation

Open questions:
- show the notifications be like toast notifications?
- how important are such notifications? should notifications be in real time?
- how often the user is logged inside the UI and can see the notifications?
- should the notification be visible inside a card in the dashboard?
- can we reuse the same system to also save the UI errors?


gsanchietti:
Not needed for now.
I'm going to keep the branch for future use.

PR #347, opened 2024-02-20
ns-api: readme, fix netmap example

gsanchietti:
The example was wrong.

Card: https://trello.com/c/aSCKhOGD/332-netmap-ui

### DIFF
--- a/packages/ns-api/README.md
+++ b/packages/ns-api/README.md
@@ -5645,7 +5645,7 @@ Reponse example:
 
 Add new netmap rules.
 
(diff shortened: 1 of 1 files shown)


packages/ns-api/README.md
```diff
@@ -5644,12 +5644,13 @@ Reponse example:
 ### add-rule
 
 Add new netmap rules.
-Add a source rule:
+
+Add a destinarion rule:
```
andre8244, line 5648:
```suggestion
Add a destination rule:
```

PR #365, opened 2024-03-10
ns-plug: avoid sending backup of non-config files

filippocarletti:
Hack to "remove" modified non-config files from the backup.
Support a list of space separated files to skip.


gsanchietti (changes requested):
I see 3 problems:
- we just removed an hack for ACME, why adding a new one?
- /etc/banip/banip.blocklis is a local config, and should be kept inside the backup
- if we still want to keep the hack, let's move the list of file outside the script

PR #368, opened 2024-03-11
ns-dpi: download netify apps on boot

filippocarletti:
Download netify apps and categories on boot.
https://trello.com/c/tWClJOWa


gsanchietti:
Fixed indentation.

PR #373, opened 2024-03-15
ns-migration: network, fix alias

gsanchietti:
Changes:

- refactor alias alias format to match the one required by the UI
- create the device records for the ethernet interfaces: this is also required by the UI
- refactor bond format to match the one required by the UI
- show inside the UI the bond slaves and the bridge ports

The export archive now must contain the 'devices' node inside network.json file
Minimum release of exporter: nethserver-firewall-migration-0.0.12-1.ns7.noarch.rpm

Card: https://trello.com/c/Xxqd9mFS/372-migration-alias-ips
Card: https://trello.com/c/DREXWyEM/377-migration-remapping-should-display-only-physical-devices

### DIFF
--- a/packages/ns-api/files/ns.dhcp
+++ b/packages/ns-api/files/ns.dhcp
@@ -68,8 +68,8 @@ def list_interfaces():
         dhcps = []
     for i in interfaces:
         record = {"device": "", "start": "", "end": "", "active": False, "options": {} }
-        # skip loopback, bond devices, wans and interfaces with no IP address (e.g. DHCP interfaces)
-        if i == "loopback" or i in wans or not ('ipaddr' in interfaces[i]) or interfaces[i].get('proto') == 'bonding':
+        # skip loopback, bond devices, wans, aliases and interfaces with no IP address (e.g. DHCP interfaces)
+        if i == "loopback" or i in wans or not ('ipaddr' in interfaces[i]) or interfaces[i].get('proto') == 'bonding' or interfaces[i].get('device', '').startswith('@'):
             continue
 
         record['device'] = interfaces[i].get('device', '')
--- a/packages/ns-api/files/ns.migration
+++ b/packages/ns-api/files/ns.migration
@@ -55,6 +55,8 @@ elif cmd == 'call':
 
     elif action == 'list-source-devices':
         ret = []
+        bridges = {}
+        devices = {}
         try:
             # read input and prepare the paths
             data = json.load(sys.stdin)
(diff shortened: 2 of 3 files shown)


cotosso:
Definitely better than the previous image, but there is still something to fix:
**1.**
lan1 ha 2 aliases, one is displayed near the lan1 ipaddress, the other one is displayed like an alias 
image

This is because the code creates one interface for every alias 
```
config interface 'eth0_2'
        option device '@lan1'
        option proto 'static'
        list ipaddr '192.168.1.2/24'

config interface 'eth0_1'
        option device '@lan1'
        option proto 'static'
        list ipaddr '192.168.1.3/24'

```

instead we need to create only 1 interface with all aliases listed, like this 
```
config interface 'eth0_1'
        option device '@lan1'
        option proto 'static'
       list ipaddr '192.168.1.2/24'
        list ipaddr '192.168.1.3/24'
```

**2.**

Devices have been doubled in the network db: in this extract (taken from the db) device eth0 appears twice as well as eth1, device vlan10 appears twice as well.

```
config device 'ns_f582c3f8'
        option name 'eth0'
        option ipv6 '0'

config device 'ns_d18231c1'
        option name 'eth1'
        option ipv6 '0'

config device 'ns_5150d732'
        option name 'eth0'
        option ipv6 '0'

config device 'ns_77482e8e'
        option name 'eth1'
        option ipv6 '0'

config device 'vlan10'
        option type '8021q'
        option vid '10'
        option ifname 'eth0'
        option name 'eth0.10'

config device 'vlan10a22f4'
        option type '8021q'
        option vid '10'
        option ifname 'eth0'
        option name 'eth0.10'

```

cotosso:
Latest changes fixed previous problems with aliases: multiple aliases well managed and no duplicated interfaces:
image

The only detail I noticed was the variable pattern for **name** of the alias interface created, sometimes there is **al_** at the start , sometimes not,  but it's **only** a detail.
image

cotosso:
new test with bond interfaces: there is a problem on aliases of bonds.
this is an extract fron the migrated db:

```
config interface 'bond0'
        option proto 'bonding'
        option bonding_policy 'active-backup'
        list slaves 'eth2'
        list slaves 'eth3'
        option link_monitoring 'mii'
        option miimon '100'
        option netmask '255.255.255.255'
        option ipaddr '127.180.79.1'

config interface 'bond0_lan'
        option device 'bond-bond0'
        option ipaddr '10.20.20.20'
        option proto 'static'
        option netmask '255.255.255.0'


config interface 'al_bond05'
        option device '@bond0'
        list ipaddr '10.20.20.21/24'
        option proto 'static'
```

the alias interface must be an alias of the configured interface *bond0_lan* and not of the *bond0* interface
using a "diff-like" syntax it should be:

```
config interface 'al_bond05'
-        option device '@bond0'
+        option device '@bond0_lan'
        list ipaddr '10.20.20.21/24'
        option proto 'static'
```

Issue #417, opened 2024-04-15
Let’s Encrypt not working


gsanchietti:
**Steps to reproduce**

- Access the Certificates page
- Request a Let's Encrypt certificate

**Expected behavior**

Let's Encrypt certificate installation should complete successfully without any errors.
After completion, the user should have a valid SSL certificate installed on their domain(s).

**Actual behavior**

All Let's Encrypt certificates remain in pending status.

Execute:
```
/etc/init.d/acme restart
```

Reported error:
```
acme: Option "standalone" is deprecated.
acme: Please set "option validation_method standalone".
/usr/lib/acme/hook: line 47: keylength: parameter not set
```

**Components**

Version: NethSecurity 8 23.05.3-ns.0.0.3-rc1

**See also**

https://community.nethserver.org/t/lets-encrypt-not-working/23326

----

Thanks to **rstedmond **

gsanchietti:
Packages:
- ns-api-0.0.61
- acme-common-1.1.2

PR #418, opened 2024-04-15
Fix Let's Encrypt certificate request

gsanchietti:
#417


Tbaile (approved):
Love the approach :see_no_evil:

Issue #420, opened 2024-04-15
FlahStart: redirect rule is ineffective


gsanchietti:
**Steps to reproduce**

- Access the "FlashStart" page
- Enable the service on LAN zone and add the required credentials
- On FlashStart remote portal, setup a custom blacklist; eg. block the domain nethsecurity.org

**Expected behavior**

All DNS queries from LAN clients must be redirect to FlashStart DNS: query to nethsecurity.org must return FlashShstart address.

**Actual behavior**

DNS queries are not redirected to FlashStart.

**Components**

Version: 8-23.05.3-ns.0.0.3-rc1
Package: ns-flashstart - 0.0.3-1

gsanchietti:
Version: 23.05.3-ns.0.0.3-rc1-35-g0721453

Packages:
- ns-flashstart-0.0.4
- ns-api-0.0.63

PR #421, opened 2024-04-15
ns-flashstart: fix firewall rules

gsanchietti:
Changes:
- fix redirect rules
- force dnsmasq restart

See:
- #420 
- #422


cotosso:
After testing everyting seems to work as expected:
* redirect is effective on selected zones
* bypass are effective too
* after disabling flashstart dns service works flawlessy

PR #433, opened 2024-04-17
packages: add irqbalance

filippocarletti:
irqbalance should improve network performance on multi-core systems.
Decision about default installation postponed after benchmarks.

```
opkg update
opkg install irqbalance
uci set irqbalance.irqbalance.enabled=1
uci commit
service irqbalance start
```

### DIFF
--- a/config/irqbalance.conf
+++ b/config/irqbalance.conf
@@ -1,0 +1,1 @@
+CONFIG_PACKAGE_irqbalance=y


config/irqbalance.conf
```diff
@@ -0,0 +1 @@
+CONFIG_PACKAGE_irqbalance=y
```
gsanchietti, line 1: Build it as a module:
```suggestion
CONFIG_PACKAGE_irqbalance=m
```

gsanchietti (changes requested):

filippocarletti:
Benchmarks on a 2 cores (4 threads) i5-3320M do NOT show relevant performance gain.
Tested with iperf3 and curl.

filippocarletti:
Also on i7. Closing.

PR #439, opened 2024-04-18
conf: add strongswan-mod-unity

gsanchietti:
The plugin should enable to route multiple networks on the same tunnel.

This is a test requested by @cotosso


gsanchietti:
Not required anymore.

Issue #471, opened 2024-04-30
Threat shield: remove enterprise list on subscription removal


gsanchietti:
Yoroi and Nethesis lists are available only with a valid subscription and a Threat Shield entitlement.

Add an hook to remove the Yoroi/Nethesis lists when the subscription has been removed.

andre8244:
Yoroi/Nethesis lists are removed as expected after subscription cancellation

PR #472, opened 2024-04-30
ns-threat_shield: remove premium lists on unregister

gsanchietti:
When the machine is unregistered, the premium lists are not accessible anymore.

#471

### DIFF
--- a/packages/ns-threat_shield/files/ts-ip
+++ b/packages/ns-threat_shield/files/ts-ip
@@ -9,6 +9,18 @@
 # Threat shield: add Nethesis categories to banip
 #
 
+. /lib/functions.sh
+
+keep=''
+
+handle_blocklist() {
+    # preserve only non-enteprise lists
+    if ! echo $1 | grep -q -E '^yoroi|^nethesis' ; then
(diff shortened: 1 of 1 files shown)


packages/ns-threat_shield/files/ts-ip
```diff
@@ -9,6 +9,19 @@
...
 #
 
+. /lib/functions.sh
+
+keep=''
+
+handle_blocklist() {
+    local value="$1"
+    # preserve only non-enteprise lists
+    if ! echo $1 | grep -q -E '^yoroi|^nethesis' ; then
```
filippocarletti, line 19:
```suggestion
    if ! echo $value | grep -q -E '^yoroi|^nethesis' ; then
```

filippocarletti (commented):
Or remove line 17.

packages/ns-threat_shield/files/ts-ip
gsanchietti: Let's remove the line 17

filippocarletti (approved):

PR #486, opened 2024-05-03
ts-ip: fix gawk regexp

filippocarletti:
The previous regexp gives the following error:
gawk: cmd. line:1: warning: regexp escape sequence `\d' is not a known regexp operator
and the feed is ignored:
May  3 17:09:47 fw banIP-0.9.4-3[16535]: skip empty feed 'yoroisusplvl2v4'


gsanchietti:
Please open an issue and the describe how to reproduce.

filippocarletti:
https://github.com/NethServer/nethsecurity/issues/496

gsanchietti (approved):
The new regex is much similar to the ones existing inside the banip default feed :+1:

Issue #412, opened 2024-04-11
Repository access: add a subscription authentication proxy


gsanchietti:
Access to subscription repository channel should be enabled only after the system_key has been verified.

gsanchietti:
Server implementation: https://github.com/nethesis/parceler

Tbaile:
TEST CASES:
- [ ] Fresh Install, activate **enterprise** subscription, check if there are updates, on success test case is passed.
- [ ] Fresh Install, activate **community** subscription, check if there are updates, on success test case is passed.
- [ ] From old, **already activated enterprise**, update and check for updates, on success test case is passed.
- [ ] From old, **already activated community**, update and check for updates, on success test case is passed.
- [ ] From unactivated installation, update and activate **enterprise** subscription, check if there are updates, on success test case is passed.
- [ ] From unactivated installation, update and activate **community** subscription, check if there are updates, on success test case is passed.

gsanchietti:
Testing image: 23.05.3-ns.1.0.1-16-g36f10a3

gsanchietti:
Testing image:  23.05.3-ns.1.0.1-17-gb61b927

gsanchietti:
Test 5 and 6 verified.

PR #504, opened 2024-05-08
ns-plug: added enterprise distfeed

Tbaile:
Ref: #412


packages/ns-plug/files/distfeed-setup
```diff
@@ -5,15 +5,25 @@
 # SPDX-License-Identifier: GPL-2.0-only
 #
 
-base_url=${BASE_URL:='https://updates.nethsecurity.nethserver.org'}
+base_url=${BASE_URL:=''}
```
gsanchietti, line 8: We could totally remove this line

Tbaile: Dunno what scripts actually use the manual env set of `BASE_URL`, left it there just in case :shrug:

gsanchietti: we can just remove the support of env var
```suggestion
base_url="$(uci -q get ns-plug.config.repository_url)"
```

packages/ns-plug/files/distfeed-setup
```diff
@@ -5,24 +5,27 @@
...
 channel=${CHANNEL:=''}
 owrt_version=${OWRT_VERSION:=''}
 
+if [ -z "$base_url" ]; then
+    base_url="$(uci -q get ns-plug.config.repository_url)"
+fi
+
```
gsanchietti, line 15: Not needed anymore if initialized on top.
```suggestion
```

packages/ns-api/files/ns.update
```diff
@@ -86,11 +87,15 @@ def install_package_updates():
     return {"result": "success"}
 
 def check_system_update():
+    e_uci = EUci()
     current_version = get_system_version()
     ret = {"currentVersion": f'NethSecurity {current_version}', "lastVersion": "", "scheduledAt": get_update_schedule()}
     try:
-        channel = get_distfeed_channel()
-        with urllib.request.urlopen(f"https://updates.nethsecurity.nethserver.org/{channel}/latest_release") as resp:
+        url = f"{e_uci.get('ns-plug', 'config', 'repository_url')}"
```
gsanchietti, line 94: No need to cast to string:
```suggestion
        url = e_uci.get('ns-plug', 'config', 'repository_url', default="https://updates.nethsecurity.nethserver.org")
```

gsanchietti (commented):
Also add default repository_url to https://github.com/NethServer/nethsecurity/blob/main/packages/ns-plug/files/config

gsanchietti:
Also, for the HTTP basic authentication take a look at this: https://stackoverflow.com/questions/2407126/python-urllib2-basic-auth-problem

packages/ns-api/files/ns.update
Tbaile: I'd remove the default, it is set anyway, by default pyuci returns a string type
```suggestion
        url = e_uci.get('ns-plug', 'config', 'repository_url')
```

gsanchietti: ok remove the default and the f-string

packages/ns-plug/files/register
```diff
@@ -76,6 +76,8 @@ esac
 
 uci set ns-plug.config.system_id="$system_id"
 uci set ns-plug.config.secret="$secret"
+uci set ns-plug.config.repository_url="https://$system_id:$secret@distfeed.nethesis.it/repository/$type/nethsecurity"
+
```
gsanchietti, line 80:
```suggestion
```

files/etc/uci-defaults/99-nethsec-default-distfeed
```diff
@@ -0,0 +1,6 @@
+[ "$(uci -q get ns-plug.config.repository_url)" != "" ] && exit 0
+
+uci -q batch << EOI
+set ns-plug.config.repository_url="https://updates.nethsecurity.nethserver.org"
+commit ns-plug
+EOI
```
gsanchietti, line 6: Move the implementation inside [20_ns-plug script](https://github.com/NethServer/nethsecurity/blob/main/packages/ns-plug/files/20_ns-plug), before line 9:
- the url must be set even when updating with packages, the 20_ns-plug is [already executed on update](https://github.com/NethServer/nethsecurity/blob/main/packages/ns-plug/Makefile#L40-L48)
- the url must be set before invoking distfeed

gsanchietti (changes requested):

gsanchietti (approved):

packages/ns-plug/files/20_ns-plug
```diff
@@ -5,5 +5,17 @@ crontab -l | grep -q '/usr/sbin/send-backup' || echo '02 2 * * * sleep $(( RANDO
...
+if [ "$(uci -q get ns-plug.config.repository_url)" = "" ]; then
+    if [ "$(uci -q get ns-plug.config.system_id)" != "" ]; then
+        type="$(uci -q get ns-plug.config.type)"
+        system_id="$(uci -q get ns-plug.config.system_id)"
+        secret="$(uci -q get ns-plug.config.secret)"
+        uci set ns-plug.config.repository_url="https://$system_id:$secret@distfeed.nethesis.it/repository/$type/nethsecurity"
+    else
+        uci set ns-plug.config.repository_url="https://updates.nethsecurity.nethserver.org"
+    fi
+    uci commit
```
gsanchietti, line 17:
```suggestion
    uci commit ns-plug
```

gsanchietti (changes requested):

gsanchietti (approved):

Issue #505, opened 2024-05-08
Threat shield: blocklist not applied to all WANs


gsanchietti:
There is an issue with the autodetection feature in BanIP, especially in the presence of multi-WAN setups. The autodetection of WAN interfaces does not always work correctly.

**Steps to reproduce**

1. Configure BanIP on a machine with multi-WAN enabled.

**Expected behavior**

The autodetection feature should accurately identify all WAN interfaces and devices used for monitoring, ensuring that blocklists are applied correctly on all WANs.

**Actual behavior**

Autodetection may fail to identify all WAN interfaces and devices accurately, leading to incomplete blocklist application on certain WANs.

Check wan-input banip chain:
```
# nft list chain inet banIP wan-input
table inet banIP {
	chain wan-input {
		type filter hook input priority -200; policy accept;
                ...
		iifname != "eth2" counter packets 31 bytes 2232 accept
	       ...
	}
}
```

The line containing the `ifname` directive should check all existing wans.

**Components**

NethSecurity version: 8-23.05.3-ns.0.0.5-rc2-9-gb67b1d3

cotosso:
Almost everything perfect.
Now banip configuration is correctly created based on interfaces configuration, I created and removed many times WAN interfaces and the config file is alway correct:

```
root@NethSec-Z1:~# uci show banip
banip.global=banip
banip.global.ban_enabled='1'
banip.global.ban_debug='0'
banip.global.ban_autodetect='0'
banip.global.ban_logterm='Exit before auth from' 'luci: failed login' 'error: maximum authentication attempts exceeded' 'sshd.*Connection closed by.*\[preauth\]' 'SecurityEvent=\"InvalidAccountID\".*RemoteAddress=' 'received a suspicious remote IP '\''.*'\'''
banip.global.ban_fetchcmd='curl'
banip.global.ban_protov4='1'
banip.global.ban_ifv4='WAN1' 'WAN2' 'WAN3'
banip.global.ban_ifv6='WAN1' 'WAN2' 'WAN3'
banip.global.ban_trigger='WAN1' 'WAN2' 'WAN3'
banip.global.ban_dev='eth1' 'eth2' 'eth3'
```

Unfortunately **nft rules** are **not** updated after WAN creation/deletion, this means that this section:

```
      chain wan-input {
.
.
                iifname != { "eth1", "eth3" } counter packets 0 bytes 0 accept
```

remains the same after WAN creation/deletion even if banip database has changed.

[...]

PR #521, opened 2024-05-13
ns-api: banip, set missing options

gsanchietti:
Make sure to set the options that are usually controlled by 'autodetect' which is now disabled.

#505


andre8244 (approved):
:man_shrugging:

Issue #492, opened 2024-05-07
Install from USB: bad partition table


gsanchietti:
**Description**

The partition table generated from a USB pen written by Windows is incorrect. 
This issue arises specifically when creating the USB pen for NethSecurity 8.
Upon inspection, it was found that when the USB pen is created on Windows, three partitions with IDs 128/1/2 are reported, whereas on Linux with `dd`, this process occurs correctly. 
However, on Windows, whether using tools like Balena, Win32, or Rufus, the partitions are initially written correctly, but upon completion of the writing process, when Windows mounts the pen, it performs a check and changes the ID 128 to 3 because it doesn't like the sequence. 
Consequently, when performing an `ns-install`, the firewall is created with IDs 3-1-2. This causes issues when enabling logging, as the ID 3 is already occupied, disrupting the functionality.

**Steps to reproduce**

1. Create a USB pen for NethSecurity 8 on Windows using tools like Balena, Win32, or Rufus.
2. Observe the partition table generated on the USB pen.

**Expected behavior**

The partition table should be correctly generated without any alterations to the partition IDs.

**Actual behavior**

After creating the USB pen on Windows, the partition table displays IDs 128/1/2, but upon mounting on Windows, the ID 128 is changed to 3, causing issues during the installation process.

If ns-install is used on a legacy-BIOS machine, the only effect of the bug is a wrong partition numbering that leads to a bug on the extra storage management: the free disk partition cannot be used for extra storage due to the incorrect partition table.

If ns-install is used on UEFI machine, the installation fails with the error:
```
/usr/sbin/ns-install: line 36: -1: substring expression < 0
```
This is because the `/boot` partition is not mounted.

**Additional Information**

Bad partition table:
```
Number  Start   End     Size    File system  Name  Flags
        17.4kB  262kB   245kB   Free Space
 1      262kB   17.0MB  16.8MB  fat16              legacy_boot
 2      17.0MB  332MB   315MB
        332MB   64.0GB  63.7GB  Free Space
```

**Components**

NethSecurity version: 8-23.05.3-ns.0.0.3-rc1-80-g615c01e

----

Thanks to **lucad** and @NethNick for reporting.

filippocarletti:
I'd prefer to use non-EFI images, the partition table has only two partitions and the filesystems are unreadable by Windows.

gsanchietti:
Changing the name of the image may have a big impact on the following:
- build system
- CI: publication to the repository of created images
[...]

PR #525, opened 2024-05-13
Update ns-install

NethNick:
Updated for EFI reordered partitions

#492


gsanchietti:
Manually merged inside #519

PR #534, opened 2024-05-16
ns-phonehome: fix json format

gsanchietti:
The `distro` node is required, otherwise the server will not accept the payload.

Regression introduced in commit 2b9e1d1251924562d90ec8a8514b70d816519f05

Server error:
```
n1 app[1492815]: [2024-05-16 12:49:38] production.DEBUG: JsonSchema validation failed {"/facts":["The required properties (distro) are missing"]}
```

### DIFF
--- a/packages/ns-phonehome/files/phonehome
+++ b/packages/ns-phonehome/files/phonehome
@@ -74,10 +74,13 @@ if not product:
     except:
         product = ""
 
+nethsec_version = ""
+version = ""
 with open('/etc/os-release', 'r') as file:
     for line in file:
-        if line.startswith("VERSION_ID"):
+        if line.startswith("VERSION_ID="):
             nethsec_version = line.split('=')[1].split('-')[2][3:]
+            version = line.split('=')[1].replace('"','').rstrip()
             break
 
 data = {
@@ -85,6 +88,10 @@ data = {
     "uuid": sid,
     "installation": "nethsecurity",
     "facts": {
+        "distro": {
+            "name": "NethSecurity",
+            "version": version
+        },
         "processors": { 
             "count": _run("grep processor /proc/cpuinfo  | wc -l"), 
             "model":  _get_cpu_field("Model name", cpu_info),


packages/ns-phonehome/files/phonehome
```diff
@@ -74,17 +74,25 @@ if not product:
     except:
         product = ""
 
+nethsec_version = ""
+version = ""
 with open('/etc/os-release', 'r') as file:
     for line in file:
```
Tbaile, line 80:
```suggestion
with open('/etc/os-release', 'r') as file:
    if nethsec_version != "" and version != "":
        break
    for line in file:
```

packages/ns-phonehome/files/phonehome
```diff
@@ -74,17 +74,25 @@ if not product:
...
-        if line.startswith("VERSION_ID"):
+        if line.startswith("VERSION_ID="):
             nethsec_version = line.split('=')[1].split('-')[2][3:]
             break
```
Tbaile, line 83:
```suggestion
            
```

packages/ns-phonehome/files/phonehome
Tbaile, line 1: Changes requested there to avoid a empty `version` field in case a `nethsec_version` is found and the `break` is called, exiting the loop

Tbaile (changes requested):

packages/ns-phonehome/files/phonehome
```diff
@@ -74,17 +74,25 @@ if not product:
...
 
 data = {
     "$schema": "https://schema.nethserver.org/facts/2022-12.json",
     "uuid": sid,
     "installation": "nethsecurity",
     "facts": {
+        "distro": {
+            "name": "NethSecurity",
+            "version": version
+        },
```
Tbaile, line 95: No clue how this was missing, probably a wrongfully CRTL+X, ty :pray:

packages/ns-phonehome/files/phonehome
gsanchietti: Well, it works already because VERSION comes before VERSION_ID in os_release file, but I agree to fix it.

Tbaile: I know how the `os-release` file is made, however we've seen broken configuration files in packages from upstream, everything can happen :upside_down_face:

Tbaile (approved):

Issue #538, opened 2024-05-17
OpenVPN Road Warrior: OTP authentication, VPN disconnects after one hour


gsanchietti:
**Steps to reproduce**

- Configure a Road Warrior OpenVPN server with "Username and OTP authentication"
- Create a VPN user
- Connect the VPN user
- Observe the VPN connection over a period to notice intermittent disconnections.

**Expected behavior**

The VPN connection should remain stable without intermittent disconnections, as experienced in the previous version.

**Actual behavior**

Users are experiencing frequent VPN disconnections, requiring manual reactivation.

Upon further investigation, setting reneg_sec to 0 on both the server and client sides seems to mitigate the issue, allowing the connection to remain stable beyond the hour mark without requiring OTP reauthentication.

---

Thanks to @francio87

francio87:
Confirmed, now both the server and the client have the `reneg_sec=0` option set

PR #539, opened 2024-05-17
ns-api: ovpnrw, disable reneg-sec for OTP

gsanchietti:
When OTP authentication is enabled, the server should not request the password again after one hour.

#538


andre8244 (approved):
:man_shrugging:

PR #543, opened 2024-05-20
config: add nat helpers

filippocarletti:
Needed to allow active ftp sessions
Note: windows FTP client only uses active FTP

https://github.com/NethServer/nethsecurity/issues/544

### DIFF
--- a/config/nat_helpers.conf
+++ b/config/nat_helpers.conf
@@ -1,1 +1,4 @@
 CONFIG_PACKAGE_kmod-nf-nathelper=y
+CONFIG_PACKAGE_kmod-asn1-decoder=y
+CONFIG_PACKAGE_kmod-lib-textsearch=y
+CONFIG_PACKAGE_kmod-nf-nathelper-extra=y
--- a/config/nat_helpers.conf
+++ b/config/nat_helpers.conf
@@ -1,0 +1,1 @@
+CONFIG_PACKAGE_kmod-nf-nathelper=y
--- a/config/sipalg.conf
+++ b/config/sipalg.conf
@@ -1,3 +1,0 @@
-CONFIG_PACKAGE_kmod-asn1-decoder=y
-CONFIG_PACKAGE_kmod-lib-textsearch=y
-CONFIG_PACKAGE_kmod-nf-nathelper-extra=m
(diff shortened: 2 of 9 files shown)


gsanchietti (commented):
I'm not sure this is the right approach: we already have other helpers that must be explicitly loaded, see https://dev.nethsecurity.org/design/nat_helpers/

I'd prefer to keep the same behavior for all nat helpers.

filippocarletti:
You can tell that ftp is special because it has a package only for itself.

gsanchietti:
> I'd agree, but the previous version loaded the FTP helper by default for a reason. Upgrade's users will need to read the docs and enable this single helper.

The previous release was loading all available helpers, including the ones that now are not loaded like sip-alg.
This change is already documented: https://docs.nethsecurity.org/en/latest/migration.html#migrated-configurations
So, in case of migration, users will already need to fix NAT helpers configuration.

I propose to change the behavior of all NAT helpers:
- install them by default, remove the package since is almost impossible to restore it in case of upgrades
- disable all helpers by default
- change the documentation to reflect the new behavior

> You can tell that ftp is special because it has a package only for itself.

I know that having the ftp module loaded by default should not harm, but I still prefer to have the same configuration for all NAT helpers.

gsanchietti:
@filippocarletti could u take a look?

cotosso:
I'm afraid that not having the ftp helper enabled by default will cause a lot of support requests and give the wrong impression of a product that doesn't work properly.
Since it is still a widely used protocol I would prefer to make an exception for ftp.

gsanchietti:
> I'm afraid that not having the ftp helper enabled by default will cause a lot of support requests and give the wrong impression of a product that doesn't work properly. Since it is still a widely used protocol I would prefer to make an exception for ftp.

In NethServer 7, having NAT helpers enabled by default caused many headache in the past.
 
Having the same behavior for all helpers is easier to maintain, explain and support.
When the NAT helpers will have their own page inside the UI, the behavior will be straightforward.

cotosso:
> When the NAT helpers will have their own page inside the UI, the behavior will be straightforward.

In this case I agree with you @gsanchietti

PR #549, opened 2024-05-27
ns-phonehome: removing version from facts

Tbaile:
~~Opened now the latest version, it appears that the version is uncorrectly stripped out of endline and quotes (from /etc/os-release)~~

~~This fix works with both latest versions and testing ones~~

Removed version definition in facts, for Nethsecurity expecting to use version inside distro object.


Tbaile:
Server fix ref: Nethserver/phonehome#116

gsanchietti (approved):

PR #574, opened 2024-06-05
files: override banip feeds

gsanchietti:
Some feeds have changed their own URL.
This package overrides existing feed list without
updating the whole banip.

See https://community.nethserver.org/t/download-for-feed-failed/23750


gsanchietti:
This is not a good solution because the file can't be overridden with a normal package update.

PR #586, opened 2024-06-12
Fix validation error in ns-api code

stephdl:
This pull request fixes the validation message to  `passphrase`. This PR updates the code to correctly handle the validation and provides a more informative error message inside the UI

REFs:
- https://github.com/NethServer/nethsecurity-ui/pull/295
- #587

### DIFF
--- a/packages/ns-api/files/ns.backup
+++ b/packages/ns-api/files/ns.backup
@@ -129,7 +129,7 @@ elif cmd == 'call':
             # reboot takes a few seconds to complete, enough to send the response
             print(json.dumps({'message': 'success'}))
         except RuntimeError as error:
-            print(json.dumps(utils.generic_error(error.args[0])))
+            print(json.dumps(utils.validation_error('passphrase',error.args[0])))
 
     elif action == 'set-passphrase':
         try:


packages/ns-api/files/ns.backup
```diff
@@ -129,7 +129,7 @@ elif cmd == 'call':
             # reboot takes a few seconds to complete, enough to send the response
             print(json.dumps({'message': 'success'}))
         except RuntimeError as error:
-            print(json.dumps(utils.generic_error(error.args[0])))
+            print(json.dumps(utils.validation_error('passphraseRef',error.args[0] )))
```
andre8244, line 132: - `restore failed` is a runtime error and should be returned to the UI as `generic_error`
- `decryption failed` is raised when the backup file is invalid or the passphrase is wrong (more likely): let's always handle it as a validation error
- Let's put a `if` to check the error string contained in the runtime error to return to the UI a generic error for `restore failed` and a validation error for `decrypt failed`
- `ref`s are a Vue concept, let's use `passphrase` instead of `passphraseRef`

Tbaile: There are 2 exceptions that can be catched:
- `decryption failed` coming from `restore_backup`
- `restore failed` still coming from `restore_backup`

I'd suggest the following (and Giacomo is gonna hate me)
- create two different exceptions in the `restore_backup`, extending the RuntimeException one (just a different name will suffice)
- return `generic_error` or `validation error` using `except` (while checking the content of an exception is fine, you will never know that the exception might contain...)
- add a final `except Exception` to catch an unforseen error coming from wherever.

Tbaile: Don't agree, if the user puts in a random file the issue is not the passphrase, is the file

Tbaile: Okay, final decision from the PM, use the `validation_error` method to send the error to validate to the user
I'd stick with what Andrea said, just a `passphrase` will suffice
Everything will be taken care of from frontend side

packages/ns-api/files/ns.backup
```diff
@@ -129,7 +129,7 @@ elif cmd == 'call':
             # reboot takes a few seconds to complete, enough to send the response
             print(json.dumps({'message': 'success'}))
         except RuntimeError as error:
-            print(json.dumps(utils.generic_error(error.args[0])))
+            print(json.dumps(utils.validation_error('passphrase',error.args[0] )))
```
andre8244, line 132:
```suggestion
            print(json.dumps(utils.validation_error('passphrase',error.args[0])))
```

andre8244 (approved):
Let's remove a space char and then it's ok

Tbaile (approved):
looks fine, remember to squash the changes and follow commit scheming [provided by the project and openwrt](https://dev.nethsecurity.org/development_process/#commit-message-style-guide):
`package-changed: commit info`

Issue #587, opened 2024-06-13
Restore: show a validation error for invalid passphrase


andre8244:
Show a validation error "Invalid passphrase" if the user enters a wrong passphrase when restoring a backup.
At the moment the standard toast notification for API failed is shown.

stephdl:
QA

install the test image

go to backup and restore, do a backup with password

try to restore it, the error form of validation must be different if you give no password or if you give a wrong password

gsanchietti:
Testing image: 23.05.3-ns.1.0.1-11-g7926d6e

gsanchietti:
The error is displayed correctly only when restoring a backup with an upload: verified.
If machine has a subscription, and the backup is restored from a remote backup, the old behavior is still present: not verified.

Screenshot of failed verification:
![image](https://github.com/NethServer/nethsecurity/assets/804596/b14219cc-73dd-4960-9845-89ca01603a4f)

gsanchietti:
Now the message is correctly displayed also when restoring a remote backup.

Verified.

PR #596, opened 2024-06-20
Fix validation for enterprise in ns.backup

stephdl:
This pull request fixes a validation issue in the `ns.backup` file for the enterprise feature. 

This pull request fixes also the usage of passphrase with the stored backup, previously it was `None` 

now in case of restore we handle three errors

- wrong backup type
- encrypted archive but missing passphrase : `error.backup_passphrase_passphrase_required`
- encrypted archive but wrong passphrase

translation must be added

https://github.com/NethServer/nethsecurity/issues/587

### DIFF
--- a/packages/ns-api/files/ns.backup
+++ b/packages/ns-api/files/ns.backup
@@ -177,7 +177,7 @@ elif cmd == 'call':
             subprocess.run(['/usr/sbin/remote-backup', 'download', data['file'], backup_path],
                            check=True, capture_output=True)
             # fetch completed successfully, restore system
-            restore_backup(backup_path, None)
+            restore_backup(backup_path, data.get('passphrase', None))
             # restore successful, restart the system
             subprocess.run(['/sbin/reboot'], check=True, capture_output=True)
             # reboot takes a few seconds to complete, enough to send the response
--- a/packages/ns-api/files/ns.backup
+++ b/packages/ns-api/files/ns.backup
@@ -50,9 +50,21 @@ def create_backup():
             raise RuntimeError('backup failed')
     return file_name
 
+def is_encrypted(file_path):
+    try:
+        # Run the 'file' command on the specified file
+        result = subprocess.run(['file', file_path], capture_output=True, text=True, check=True)
+        # Check if the output contains indications of PGP encryption
+        if "GPG" in result.stdout:
+            return True
+        else:
+            return False
+    except subprocess.CalledProcessError as e:
+        return False
+
 def restore_backup(archive, passphrase=None):
     try:
-        if passphrase is not None:
+        if is_encrypted(archive) and passphrase is not None:
             subprocess.run([
                 '/usr/bin/gpg',
                 '--decrypt',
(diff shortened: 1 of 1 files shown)


gsanchietti (approved):

packages/ns-api/files/ns.backup
```diff
@@ -187,7 +187,7 @@ elif cmd == 'call':
         except KeyError as error:
             print(json.dumps(utils.validation_error('file', 'required')))
         except RuntimeError as error:
-            print(json.dumps(utils.generic_error(error.args[0])))
+            print(json.dumps(utils.validation_error('passphrase',error.args[0])))
```
gsanchietti, line 190: I think this is not correct: if the backup is corrupted, and error on passphrase is returned.

gsanchietti (changes requested):

packages/ns-api/files/ns.backup
stephdl: this is what we have already done [L132 ](https://github.com/NethServer/nethsecurity/blob/86d38d08f849b1d8deca9377944baae330ab5414/packages/ns-api/files/ns.backup#L132)

it is the same accepted solution, two error messages possible

invalid backup or empty passphrase (but required)
wrong passphrase

stephdl: it comes from https://github.com/NethServer/nethsecurity/blob/86d38d08f849b1d8deca9377944baae330ab5414/packages/ns-api/files/ns.backup#L70

only two messages are sent back

gsanchietti: > this is what we have already done L132 

So we should fix also that one.

> it is the same accepted solution, two error messages possible

we can use the messages to raise the correct exception, or change the restore_backup function to raise 2 different exceptions

gsanchietti (changes requested):
A better fix is needed.

gsanchietti (commented):
It seems to me that if the line 87 raises the exception, the API still returns with a validation error about passhprase.

Am I wrong @stephdl ?

stephdl:
I hope you are wrong, the wrong passphrase is raised by the 85, the missing passphrase when required (mens encrypted) is sent back by line 79

however actually you are right, the translation `restore failed` returns a problem of `backup file or missing passphrase`

gsanchietti:
> however actually you are right, the translation `restore failed` returns a problem of `backup file or missing passphrase`

so we need to fix it

packages/ns-api/files/ns.backup
```diff
@@ -50,9 +50,21 @@ def create_backup():
             raise RuntimeError('backup failed')
     return file_name
 
+def is_encrypted(file_path):
+    try:
+        # Run the 'file' command on the specified file
+        result = subprocess.run(['file', file_path], capture_output=True, text=True, check=True)
+        # Check if the output contains indications of PGP encryption
+        if "GPG" in result.stdout:
```
gsanchietti, line 58: If for any reason the file name contains `GPG` string, this function will return true even if the file is not really encrypted.

packages/ns-api/files/ns.backup
gsanchietti: The error is still there: if the restore fails, a validation error is raised:
image

gsanchietti (changes requested):
The PR does still not solve correctly the issue, still [this commit](https://github.com/NethServer/nethsecurity/pull/596/commits/86d38d08f849b1d8deca9377944baae330ab5414) fixes a real issue

gsanchietti:
I have cherry-picked the good commit: https://github.com/NethServer/nethsecurity/commit/44bc3f2c5d1b9ea8a89e758b48a35b5a709eafc6

stephdl:
obsoleted by https://github.com/NethServer/nethsecurity/pull/598

Issue #689, opened 2024-07-30
Add objects support in Port Forward rules


andre8244:
Add objects support in Port forward rules.
These are the fields we should cover with objects:

- Destination address (only a single IP accepted)
- Restrict access from (accepts multiple IPs and/or CIDRs).

As for firewall rules we could use a radio button to choose between raw IP/CIDR and object

andre8244:
Testing image version: [8-23.05.4-ns.1.1.99-alpha1-35-g92bbbedc](https://updates.nethsecurity.nethserver.org/dev/8-23.05.4-ns.1.1.99-alpha1-35-g92bbbedc/targets/x86/64/nethsecurity-8-23.05.4-ns.1.1.99-alpha1-35-g92bbbedc-x86-64-generic-squashfs-combined-efi.img.gz)

andre8244:
**Test case 1 (No regressions)**

- Go to **Firewall > Port forward**
- Create a port forward
  - Enter an IP address as **Destination address**
  - Expand **Advanced settings**
  - Enter one or more IP addresses as **Restricted addresses**
  - Click **Add port forward**
- Ensure port forward is created successfully and its attributes are presented correctly in the table
- Edit the port forward by changing **Destination address** and **Restricted addresses**
- Ensure port forward is modified successfully and its attributes are presented correctly in the table
- Delete the port forward

**Test case 2 (Host sets and domain sets)**

- Go to **Users and objects > Objects > Host sets**
- Create a host set containing only one record
- Create a host set containing two or more records
- Go to **Domain sets** tab
- Create a domain set with only one domain
- Create a domain set with two or more domains
- Go to **Firewall > Port forward**
- Create a port forward
  - As **Destination address type** choose **Select an object**
  - Open **Destination object** combobox, ensure that:
    - The host set containing only one record is shown in the list
    - The host set containing two or more records is not shown in the list
    - The domain sets are not shown
  - Choose the host set as **Destination object**
  - Expand **Advanced settings**
  - As **Restrict access from** choose **Select an object**
  - Ensure that both domain sets are shown in the list
  - Choose a domain set as **Restricted object**
  - Click **Add port forward**
- Ensure port forward is created successfully and its attributes are presented correctly in the table
- Edit the port forward by selecting **No object** as **Restricted object**
- Ensure port forward is modified successfully and its attributes are presented correctly in the table
- Delete the port forward

**Test case 3 (DHCP reservations)**

[...]

PR #702, opened 2024-08-02
fix: ns.redirect list-object-suggestions

andre8244:
`ns.redirect list-object-suggestions` api should return domain sets only. Before this change also other objects, e.g. VPN users and DHCP reservations were returned.

Ref:
- https://github.com/NethServer/nethsecurity/issues/689


Tbaile (approved):
If @gsanchietti is okay with this, you have full authority over the changes :+1:

gsanchietti (approved):

PR #705, opened 2024-08-05
ns-api: redirects, accept dns records

andre8244:
Accept DNS records as port forward destination.

See:
- https://github.com/NethServer/nethsecurity/issues/689


gsanchietti (approved):
Maybe at this point a positive logic was better, but looks good to me.

Tbaile (approved):

PR #707, opened 2024-08-05
ns-api(mwan): adding CIDR and IPSets to SRC and DST

Tbaile:
It appears that CIDR and IPSets were left out from MWAN rule configuration

Needs:
- https://github.com/NethServer/python3-nethsec/pull/63

### DIFF
--- a/packages/ns-api/files/ns.mwan
+++ b/packages/ns-api/files/ns.mwan
@@ -11,6 +11,12 @@ import sys
 from euci import EUci
 from nethsec import mwan, utils, objects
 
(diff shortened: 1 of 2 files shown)


packages/ns-api/files/ns.mwan
```diff
@@ -11,6 +11,13 @@ import sys
 from euci import EUci
 from nethsec import mwan, utils, objects
 
+
```
gsanchietti, line 14:
```suggestion
```

gsanchietti (approved):

PR #783, opened 2024-09-24
ci: removing hardcoded MAJOR

Tbaile:
Getting the MAJOR from the latest annotated tag.

### DIFF
--- a/.github/workflows/release-stable.yml
+++ b/.github/workflows/release-stable.yml
@@ -14,9 +14,6 @@ on:
         default: false
         type: boolean
 
-env:
-  MAJOR_VERSION: 23.05.4
-
 jobs:
   sync:
     name: Moving dev packages to stable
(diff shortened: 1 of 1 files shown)


.github/workflows/release-stable.yml
```diff
@@ -30,13 +27,21 @@ jobs:
       RCLONE_CONFIG_SPACES_ENDPOINT: ams3.digitaloceanspaces.com
       RCLONE_CONFIG_SPACES_ACL: public-read
     steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 500
+      - name: Fetch tag annotations
+        run: |
+          # Fetch tag manually because fetch-tags option for checkout@v4 does not work
+          git fetch --force --tags --depth 1
```
gsanchietti, line 36: The tag could not be on the first commit, but probably will be buried somewhere inside the history.
```suggestion
          git fetch --force --tags --depth 500
```

.github/workflows/release-stable.yml
```diff
@@ -30,13 +27,21 @@ jobs:
...
+          echo "MAJOR_VERSION=$(git describe | cut -d'-' -f1)" >> $GITHUB_ENV
```
gsanchietti, line 37: Add a safeguard here: make the action fail if MAJOR_VERSION is not set

gsanchietti (changes requested):

Issue #760, opened 2024-09-16
Unable to disable legacy LuCI UI after system upgrade


gsanchietti:
**Description:**  
It is not possible to disable the legacy LuCI UI after updating the system using an image.

**Steps to Reproduce:**  
1. Enable LuCI:
   ```bash
   uci set ns-ui.config.luci_enable=1 && uci commit ns-ui && ns-ui
   ```
2. Update the system using an image.
3. Disable LuCI:
   ```bash
   uci set ns-ui.config.luci_enable=0 && uci commit ns-ui && ns-ui
   ```

**Expected Behavior:**  
LuCI should be disabled.

**Actual Behavior:**  
LuCI remains enabled. This issue occurs only after updating the machine with an image, not on a clean installation.

**Cause:**  
The issue is caused by the presence of the file: `/etc/nginx/conf.d/luci.locations`.

---

PR:
- #771

gsanchietti:
I've tested the fix and seems to work.
The LuCI configuration is correctly added and removed, verified also with `/usr/sbin/nginx -T -c /etc/nginx/uci.conf`.

Before releasing I would like to gather more inf from this community report: https://community.nethserver.org/t/unable-to-open-nethsec-ui/24561/2

gsanchietti:
The fix introduced a regression: it's not possible to disable both UI.

Non working configuration:
```
ns-ui.config=main
ns-ui.config.luci_enable='0'
ns-ui.config.nsui_enable='0'
ns-ui.config.nsui_extra_port='9090'
ns-ui.config.nsui_extra_enable='0'
```

PR #785, opened 2024-09-24
ns-ui: fix nginx include

gsanchietti:
#760

### DIFF
--- a/packages/ns-ui/files/ns-ui
+++ b/packages/ns-ui/files/ns-ui
@@ -38,13 +38,24 @@ server {
 	listen $nsui_extra_port ssl default_server;
 	listen [::]:$nsui_extra_port ssl default_server;
 	server_name _lan;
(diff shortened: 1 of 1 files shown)


packages/ns-ui/files/ns-ui
```diff
@@ -38,13 +38,24 @@ server {
...
 	server_name _lan;
-	include conf.d/*.locations;
 	ssl_certificate $crt;
 	ssl_certificate_key $key;
 	ssl_session_cache shared:SSL:32k;
 	ssl_session_timeout 64m;
 	error_log syslog:server=unix:/dev/log;
 	access_log syslog:server=unix:/dev/log;
+
+	# configure ns UI must be enabled on $nsui_extra_port
```
Tbaile, line 48: If this doesn't resolve in the value of the variable, we're golden

PR #793, opened 2024-09-25
ns-api(threathshield): added additional options to api

Tbaile:
Adding additional APIs following the new enchancements planned:
- https://github.com/orgs/NethServer/projects/10/views/2?pane=issue&itemId=80850710


packages/ns-api/README.md
```diff
@@ -5641,6 +5656,71 @@ Response example:
...
+
+The `address` field can be an IPv4/IPv6, a CIDR, a MAC or host name
+
+Response example:
+```json
+{"message": "success"}
+```
+
+It can raise the following validation errors:
+- `address_already_present` if the address is already inside the allow list
```
gsanchietti, line 5697:
```suggestion
- `address_already_present` if the address is already inside the block list
```

packages/ns-api/README.md
```diff
@@ -5641,6 +5656,71 @@ Response example:
...
+api-cli ns.threatshield delete-blocked --data '{"address": "1.2.3.4"}'
+```
+
+Response example:
+```json
+{"message": "success"}
+```
+
+It can raise the following validation errors:
+- `address_not_found` if the address is not inside the allow list
```
gsanchietti, line 5722:
```suggestion
- `address_not_found` if the address is not inside the block list
```

packages/ns-api/README.md
```diff
@@ -5641,6 +5656,71 @@ Response example:
 It can raise the following validation errors:
 - `address_not_found` if the address is not inside the allow list
 
+### list-blocked
+
+List blocked addresses:
```
gsanchietti, line 5661:
```suggestion
List blocked addresses from the local blocklist:
```

gsanchietti (approved):

gsanchietti (changes requested):
Since this update will be released with a new image, there is no need to execute the defaults during the package update procedure.

Still, since we are defining new defaults, it does not make sense to have ban_logterm that can be applied to our default configuration.
Please change ban_logterm to have sane defaults at least for:
- luci behind nginx
- nsui
- dropbear (if simple enough)

packages/ns-threat_shield/files/banip-defaults
```diff
@@ -0,0 +1,22 @@
...
+set banip.global.ban_logprerouting="0"
+set banip.global.ban_loginput="0"
+
+set banip.global.ban_loglimit="100"
+set banip.global.ban_logcount="3"
+set banip.global.ban_nftexpiry="30m"
+
+delete banip.global.ban_logterm
+add_list banip.global.ban_logterm="Exit before auth from"
+add_list banip.global.ban_logterm="luci: failed login"
```
gsanchietti, line 15: Just remove this line because it will not match:
```suggestion
```

gsanchietti: Replace this line with:
```suggestion
add_list banip.global.ban_logterm="authentication failed for user"
```

See https://github.com/NethServer/nethsecurity-api/pull/13

gsanchietti (changes requested):
Please fix the logterm values

Tbaile:
Pushed the change in the stash then forgot to apply it, my bad!

gsanchietti (approved):

PR #798, opened 2024-09-26
ns-report: add flowreport

gsanchietti:
New daemon named `flowreport` derived from `dpireport` daemon.

The daemon reads the flow status from netifyd socket then writes the list of flows inside `/var/run/flowreport/flows.json`, every 30 seconds.
Before writing the JSON file, the daemon also cleanup stale flows and limit the number of flows to 100.

See card: https://github.com/orgs/NethServer/projects/10?pane=issue&itemId=79232989

TODO:
- [ ] create an API inside `ns.report` that reads the flows.json file
- [ ] create a mockup for the UI
- [ ] implement the UI


gsanchietti:
Postponed

PR #803, opened 2024-09-27
Add supervisor and use it to start dpireport

gsanchietti:
Add supervisor to use as init system for dpireport.

Changes:
- new supervisor daemon that starts on boot
- supervisor will start all programs listed inside `/etc/supervisord.d/*.conf` files
- dpireport drops procd support and now uses supervisor to be started at boot; it will be automatically restarted on crash

#809


Tbaile (approved):
Although it seems overkill to add this, it's needed :man_shrugging:

gsanchietti:
Replaced by https://github.com/NethServer/nethsecurity/pull/851

Issue #800, opened 2024-09-27
Centralized unit update management


Tbaile:
From the Controller it **should** be possible to update the Unit seamlessly (packages and/or image).

This is due [the feature added in the controller](https://docs.nethsecurity.org/en/latest/controller.html#version-awareness) that avoids UI - API version mismatch and inherently avoids accidental misconfigurations.

Here's the list of the actions that are to be implemented in the controller:

- [x] Unit Actions Section
  - [x] Add the action to schedule an update
  - [x] Add the action to abort an update
  - [x] Add the action to reschedule an update
  - [x] _Optional_ Add the action to force pull and install the packages

- [x] Bulk Actions Section
  - [x] Add the action to update Unit Image (scheduled or now)

- [x] Add next to the version a warning sign indicating that there's an image update available for the unit alternatively a clock indicating that there's a scheduled update planned for the unit

---
PRs:
  - [x] https://github.com/NethServer/nethsecurity-controller/pull/45 
  - [X] ~~https://github.com/NethServer/nethsecurity/pull/801~~ (not needed)
  - [x] https://github.com/NethServer/ns8-nethsecurity-controller/pull/43
  - [x] https://github.com/NethServer/nethsecurity-ui/pull/370

Tbaile:
While it's not completed and merged, the changes are only needed to be made to the `controller`.
Testing is enabled doing so:
1. On a NS8 machine, install and configure the controller with
  ```bash
add-module ghcr.io/nethserver/nethsecurity-controller:unit-remote-update
  ```
2. Prepare 3 units that are `1.0.1`, `1.1.0` and `1.2.0`.
3. Connect all units to the controller, manually refresh the info for each unit, then the following must apply:
3.1: Unit `1.0.1` and `1.1.0` must show a `Update available`
3.2: Unit `1.2.0` mustn't show any labels.

QA:
1. Check if scheduling an update works, this can be done on both `1.0.1` and `1.1.0`
2. Try to cancel a scheduled upgrade
3. For one unit, launch a image update directly and wait for it to boot up, try to access the unit (careful, openWRT sessions are long, unit might still appear connected while not so.). In addition the unit won't be having any info fetched because it's a cron, a "Update available" might still be there. 
4. Check if fetching package update works. (on `1.2.0` due to a restart of `ns-api-server` this might trigger a issue, ONCE)
5. Try the "Update units" batch command.

gsanchietti:
[...]

PR #837, opened 2024-10-10
Unit remote updates fixes

Tbaile:
Creating a script that sends up to the controller updated info, allowing the unit to push info to controller if some changes are to be reported immediately.

Ref:
- https://github.com/NethServer/nethsecurity/issues/800#issuecomment-2404234577


packages/ns-plug/files/ns-plug
```diff
@@ -110,6 +110,9 @@ explicit-exit-notify 1
 remote ${host} ${port} udp
 dev tun-nsplug
 tls-client
+script-security 2
+route-delay 2
```
gsanchietti, line 114: Remove since it's not needed:
```suggestion
```

gsanchietti (approved):

PR #844, opened 2024-10-11
Update log rotation size to a dynamic value

stephdl:
This pull request adds a script to adjust manually the size of the log, the service must be restarted after by the sysadmin

```
 ns-log-size get
 ns-log-size set 104857600
```

verify 

`cat /var/etc/rsyslog.conf`

References:
- https://github.com/orgs/NethServer/projects/10/views/2?pane=issue&itemId=81705733
- mattermost talk https://mattermost.nethesis.it/nethesis/pl/sy17gdxa8tr8tj4aptwff4qxfa
- Issue: #865

### DIFF
--- a/files/usr/sbin/ns-log-size
+++ b/files/usr/sbin/ns-log-size
@@ -22,11 +22,14 @@ get_log_rotation_size() {
 
 set_log_rotation_size() {
     local new_size=$1
+    #The value 9223372036854775807 bytes is the maximum value that can be set (8589.93 GB)
+
+    # Check if the new size is a positive integer
     if [[ ! $new_size =~ ^[0-9]+$ ]]; then
         echo "Error: Size must be a positive integer."
         exit 1
     fi
-    
+
     # Check if the new size is less than the minimum required size
     if (( new_size < MIN_SIZE )); then
         echo "Error: The input size is too low. The minimum value is $MIN_SIZE."
--- a/files/usr/sbin/ns-log-size
+++ b/files/usr/sbin/ns-log-size
@@ -1,0 +1,71 @@
+#!/bin/bash
+#
+# Copyright (C) 2024 Nethesis S.r.l.
+# SPDX-License-Identifier: GPL-2.0-only
+#
+
+# Get the current log rotation size and update it in the rsyslog configuration file
+# Usage: ns-log-size {get|set <size>}
+# Example: ns-log-size get
+# Example: ns-log-size set 104857600
+# the service must be restarted after the size is changed
+
+
+CONFIG_FILE="/etc/rsyslog.conf"
+LOG_ROTATION_LINE=$(grep -E '^\$outchannel log_rotation' "$CONFIG_FILE")
+MIN_SIZE=52428800
+
+get_log_rotation_size() {
+    local size=$(echo "$LOG_ROTATION_LINE" | awk -F', ' '{print $2}')
+    echo "$size (in bytes)"
+}
+
+set_log_rotation_size() {
+    local new_size=$1
+    if [[ ! $new_size =~ ^[0-9]+$ ]]; then
+        echo "Error: Size must be a positive integer."
+        exit 1
+    fi
+    
+    # Check if the new size is less than the minimum required size
+    if (( new_size < MIN_SIZE )); then
+        echo "Error: The input size is too low. The minimum value is $MIN_SIZE."
+        exit 1
+    fi
+
+    # Full match from start to end
+    sed -i "s/\(\$outchannel log_rotation,\/var\/log\/messages,[ ]*\)[0-9]\+\([ ]*, \/usr\/sbin\/rotate-messages\)/\1$new_size\2/" "$CONFIG_FILE"
+    
+    # Verify if the value was correctly replaced
+    local updated_line
+    updated_line=$(grep -E '^\$outchannel log_rotation' "$CONFIG_FILE")
+    if echo "$updated_line" | grep -q "$new_size"; then
+        echo "Log rotation size updated successfully to $new_size (in bytes)."
+        /etc/init.d/rsyslog restart
+        if [[ $? -ne 0 ]]; then
+            echo "Failed to restart rsyslog service."
+            exit 1
+        fi
+        echo "The rsyslog service has been restarted."
+    else
+        echo "Failed to update log rotation size."
+        exit 1
+    fi
+}
+
+case $1 in
+    get)
+        get_log_rotation_size
+        ;;
+    set)
+        if [[ -z "$2" ]]; then
+            echo "Error: No size provided for set action."
+            exit 1
+        fi
+        set_log_rotation_size "$2"
+        ;;
+    *)
+        echo "Usage: $0 {get|set <size>}"
+        exit 1
+        ;;
+esac
--- a/packages/ns-storage/files/30_ns-storage
+++ b/packages/ns-storage/files/30_ns-storage
@@ -2,9 +2,6 @@
 
 [ "$(uci -q get rsyslog.ns_memory)" == "selector" ] && exit 0
 
-echo '# Rotate in-memory log if size is bigger than 50MB' >> /etc/rsyslog.conf
-echo '$outchannel log_rotation,/var/log/messages, 52428800, /usr/sbin/rotate-messages' >> /etc/rsyslog.conf
-
 uci -q import rsyslog << EOI
 config syslog 'syslog'
 	option tcp_input_port '514'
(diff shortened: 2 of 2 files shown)


stephdl:
traffic generator https://gist.github.com/stephdl/43e0f589625d192c467f227a4998d061

gsanchietti (commented):
After reviewing the implementation and considering feedback, we noticed that the current approach of setting the log rotation size to 1GB could create issues, especially on systems with limited RAM, such as machines with only 1GB of memory. For instance, when using tmpfs for storing logs, which typically allocates less than 500MB on such systems, a 1GB log size would not fit.

We propose an alternative solution:

Set the default log rotation size to a dynamic value, such as 10% of the tmpfs size.
Additionally, allow the value to be configurable via UCI, so users with specific requirements can adjust the setting as needed.
This approach ensures better flexibility and scalability across different system configurations.

packages/ns-storage/files/30_ns-storage
```diff
@@ -18,6 +37,11 @@ config syslog 'syslog'
 config selector ns_memory
 	option source '*.*'
 	option destination ':omfile:\$log_rotation'
+
+config log_rotator 'ns_log'
+	option log_file '/var/log/messages'
+	option log_size '$log_size'
+	option rotate_command '/usr/sbin/rotate-messages'
```
gsanchietti, line 44: While this seems a nice idea, we are trying to keep the package in sync with upstream which does not have such support inside the init.d.
I eventually move this to an upstream patch.

packages/ns-storage/files/30_ns-storage
```diff
@@ -1,9 +1,28 @@
...
+	echo "Error: Could not retrieve tmpfs size, defaulting log_size to 52428800" >&2
+else
+	# Calculate 10% of tmpfs size in bytes
+	ten_percent_bytes=$(( tmpfs_size * 1024 / 10 ))
+
+	# Use default log_size if the calculated value is less than 52428800
+	if [ "$ten_percent_bytes" -gt "$log_size" ]; then
+		log_size=$ten_percent_bytes
+	fi
+fi
```
gsanchietti, line 22: I really like the logic of the minimum log_size :+1: 
Still, the current implementation has a big caveats: the `30_ns-storage` uci-default is ignored during package updates and it's not present after install; an uci-default can be used only during an image upgrade.

So this is my proposal:
- keep line 3 unchanged and remove lines from 4 to 6 (see the comment below).
- keep the minimum log_size logic and write it directly to the `/etc/rsyslog.conf` file 
- create a new script named something like `ns-log-size` or similar
- the new script does not really need a config inside UCI, but we can use the `/etc/rsyslog.conf` as "database" (the file is preserved between upgrades); the script could have 2 actions:
  - get: that read the current value
  - set: set a new value

I've asked Copilot for a proof of concept (not really tested):
```bash
#!/bin/bash

CONFIG_FILE="/etc/rsyslog.conf"
LOG_ROTATION_LINE=$(grep -E '^\$outchannel log_rotation' "$CONFIG_FILE")

get_log_rotation_size() {
    echo "$LOG_ROTATION_LINE" | awk -F', ' '{print $2}'
}

set_log_rotation_size() {
    local new_size=$1
    if [[ ! $new_size =~ ^[0-9]+$ ]]; then
        echo "Error: Size must be a positive integer."
        exit 1
    fi
    sed -i -E "s/(\$outchannel log_rotation,\/var\/log\/messages, )[0-9]+(, \/usr\/sbin\/rotate-messages)/\1$new_size\2/" "$CONFIG_FILE"
}

case $1 in
    get)
        get_log_rotation_size
        ;;
    set)
        if [ -z "$2" ]; then
            echo "Error: No size provided for set action."
            exit 1
        fi
        set_log_rotation_size "$2"
        ;;
    *)
        echo "Usage: $0 {get|set <size>}"
        exit 1
        ;;
esac
```

The new script can be called from command line or, in the future, from the UI

files/usr/sbin/ns-log-size
```diff
@@ -0,0 +1,68 @@
...
+        exit 1
+    fi
+    
+    # Check if the new size is less than the minimum required size
+    if (( new_size < MIN_SIZE )); then
+        echo "Error: The input size is too low. The minimum value is $MIN_SIZE."
+        exit 1
+    fi
+    
+    echo "Current matching line: $LOG_ROTATION_LINE"
```
gsanchietti, line 36: Remove debug line.

files/usr/sbin/ns-log-size
```diff
@@ -0,0 +1,68 @@
...
+    
+    # Full match from start to end
+    sed -i "s/\(\$outchannel log_rotation,\/var\/log\/messages,[ ]*\)[0-9]\+\([ ]*, \/usr\/sbin\/rotate-messages\)/\1$new_size\2/" "$CONFIG_FILE"
+    
+    # Verify if the value was correctly replaced
+    local updated_line
+    updated_line=$(grep -E '^\$outchannel log_rotation' "$CONFIG_FILE")
+    echo "Updated matching line: $updated_line"
```
gsanchietti, line 44: Remove debug line.

files/usr/sbin/ns-log-size
```diff
@@ -0,0 +1,68 @@
...
+    if echo "$updated_line" | grep -q "$new_size"; then
+        echo "Log rotation size updated successfully to $new_size (in bytes)."
+        echo "Please restart the rsyslog service to apply the changes."
+    else
+        echo "Failed to update log rotation size."
```
gsanchietti, line 49: Exit with 1

files/usr/sbin/ns-log-size
```diff
@@ -0,0 +1,68 @@
...
+    # Full match from start to end
+    sed -i "s/\(\$outchannel log_rotation,\/var\/log\/messages,[ ]*\)[0-9]\+\([ ]*, \/usr\/sbin\/rotate-messages\)/\1$new_size\2/" "$CONFIG_FILE"
+    
+    # Verify if the value was correctly replaced
+    local updated_line
+    updated_line=$(grep -E '^\$outchannel log_rotation' "$CONFIG_FILE")
+    echo "Updated matching line: $updated_line"
+    if echo "$updated_line" | grep -q "$new_size"; then
+        echo "Log rotation size updated successfully to $new_size (in bytes)."
+        echo "Please restart the rsyslog service to apply the changes."
```
gsanchietti, line 47: Let's restart/reload it automatically

gsanchietti (changes requested):
The new dynamic default should be set also on new installation:
call the new script inside 30_ns-storage

gsanchietti (changes requested):
See previous comment: https://github.com/NethServer/nethsecurity/pull/844#pullrequestreview-2377639901

packages/ns-storage/files/30_ns-storage
```diff
@@ -2,6 +2,7 @@
 
 [ "$(uci -q get rsyslog.ns_memory)" == "selector" ] && exit 0
 
+# first write the default configuration
 echo '# Rotate in-memory log if size is bigger than 50MB' >> /etc/rsyslog.conf
```
gsanchietti, line 6: Remove the comment because it will not be updated by the ns-log-size script
```suggestion
```

packages/ns-storage/files/30_ns-storage
```diff
@@ -2,6 +2,7 @@
...
 echo '$outchannel log_rotation,/var/log/messages, 52428800, /usr/sbin/rotate-messages' >> /etc/rsyslog.conf
```
gsanchietti, line 7: Let's remove it and write the file only once
```suggestion
```

packages/ns-storage/files/30_ns-storage
```diff
@@ -21,3 +22,21 @@ config selector ns_memory
...
+if [ -z "$tmpfs_size" ] || ! echo "$tmpfs_size" | grep -qE '^[0-9]+$'; then
+	echo "Error: Could not retrieve tmpfs size, defaulting log_size to 52428800" >&2
+else
+	# Calculate 10% of tmpfs size in bytes
+	ten_percent_bytes=$(( tmpfs_size * 1024 / 10 ))
+
+	# Use default log_size if the calculated value is less than 52428800
+	if [ "$ten_percent_bytes" -gt "$log_size" ]; then
+		log_size=$ten_percent_bytes
+		/usr/sbin/ns-log-size set $log_size
```
gsanchietti, line 40: Write directly the config file, not tested:
```suggestion
		echo '$outchannel log_rotation,/var/log/messages, $log_size, /usr/sbin/rotate-messages' >> /etc/rsyslog.conf

```

stephdl:
Playing with the limit

The value `9223372036854775807` bytes is approximately 8589.93 GB and it is accepted but `9223372036854775808` overflows the  integer limit of bash

gsanchietti (approved):
Let's wait before merge: I want to release opened issues before adding this new feature.

filippocarletti (approved):

PR #850, opened 2024-10-15
ns-migration: fixing issue with dhcp import

Tbaile:
Fixing issue where migration reports wrong naming schemas


Tbaile:
Superseded by: https://github.com/NethServer/nethsecurity/pull/856

Issue #853, opened 2024-10-16
MultiWAN: set last_resort field


gsanchietti:
It was identified that the `ns_default` policy in mwan3 does not default to "default" when the uplink is down. This behavior deviates from the intended configuration during the development phase, where the "default" value was expected for such scenarios. The author of mwan3 also recommends setting `last_resort` to "default." Recent testing has confirmed the need for this option.

**Proposed solution**

The option `uci set mwan3.ns_default.last_resort='default'` should be applied to ensure that the `last_resort` value defaults to "default" when all WAN connections fail. Additionally, it is necessary to apply the fix through a `uci-defaults` script to ensure it is implemented in existing installations. Without this, the fix will only apply during migration updates.

**Alternative solutions**

An alternative would be to maintain the current configuration, where the `last_resort` value remains set to 'unreachable'. However, this would risk no fallback mechanism if all WANs go down, preventing the automatic restoration of the connection.

**Additional context**

The issue emerged during testing, where it was observed that the `last_resort` policy had not been properly configured. The last_resort options is already set during [migration](https://github.com/NethServer/nethsecurity/commit/c8bc8237e1666dbde5bd52e03a27c7b61d77e73f).

Tbaile:
Image: TBA

QA:
1. Image update
  1.1 Install a stable version with 2 REDs
  1.2 Add 2 MWANs (you can use the "custom" type to add a single RED per policy)
  1.3 Update the unit with the image provided above
  1.4 Check through shell that `uci show mwan3 | grep last_resort` shows both MWANs and they're both set to "default"
2. Add a new MWAN
  2.1 Install the image provided above
  2.2 Add a new MWAN
  2.3 Ensure that in the "changes" section `last_resort` is set to default

cotosso:
Test Case 1 verified:
```
root@NSec8-VM-davidem:~# uci show mwan3 | grep last_resort
mwan3.ns_default.last_resort='default'
```

PR #854, opened 2024-10-16
fix: setting default mwan last_resort

Tbaile:
Ref:
- https://github.com/NethServer/nethsecurity/issues/853

### DIFF
--- a/files/etc/uci-defaults/99-nethsec-mwan
+++ b/files/etc/uci-defaults/99-nethsec-mwan
@@ -1,0 +1,5 @@
+# If `mwan3.ns_default` has no last_resort option, add it
+if [ -z "$(uci -q get mwan3.ns_default.last_resort)" ]; then
+    uci set mwan3.ns_default.last_resort=default
+    uci commit mwan3
+fi
--- a/packages/python3-nethsec/Makefile
+++ b/packages/python3-nethsec/Makefile
@@ -6,7 +6,7 @@
 include $(TOPDIR)/rules.mk
  
 PKG_NAME:=python3-nethsec
-PKG_VERSION:=0.0.79
+PKG_VERSION:=0.0.80
 PKG_RELEASE:=1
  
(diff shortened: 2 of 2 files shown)


files/etc/uci-defaults/99-nethsec-mwan
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+# If `mwan3.ns_default` is defined, add `last_resort=default` to it
+if uci -q show mwan3.ns_default > /dev/null; then
```
gsanchietti, line 4: This will override custom values.
Let's set it, only if not already set
Something like:
```
if [ -z "$(uci -q get mwan3.ns_default.last_resort)" ]; then
    uci set mwan3.ns_default.last_resort=default
fi
```

Also I think the double quotes after the `set` keyword is not required.

files/etc/uci-defaults/99-nethsec-mwan
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
```
gsanchietti, line 1: Not all uci defaults have the the shebang: not sure if is needed.

files/etc/uci-defaults/99-nethsec-mwan
```diff
@@ -0,0 +1,7 @@
...
+
+# If `mwan3.ns_default` is defined, add `last_resort=default` to it
```
gsanchietti, line 3: The comment is lying.
```suggestion
# If `mwan3.ns_default` has no last_resort option, add it
```

gsanchietti (commented):
Almost ;)

files/etc/uci-defaults/99-nethsec-mwan
Tbaile: Just to state what the file is about, uci-defaults run on `sh` anyway

gsanchietti (approved):

PR #873, opened 2024-10-24
ns-api: fix hardware info for ARM boards

gsanchietti:
On ARM boards, some devices are not present

See https://community.nethserver.org/t/nethsecurity-install-on-raspberry-pi-4b/24810


smetroid (approved):
I've tested this update, and it fixes the dashboard problem encountered during my testing

Issue #879, opened 2024-10-30
VPN Tunnel name conflict with existing interface


cotosso:
- Step 1: Create a VPN tunnel using **the same name** as an existing interface (e.g., lan1, wan).
- Step 2: Observe the configuration of the existing interface.

**Expected behavior**
The VPN tunnel should be created without affecting the settings of the existing interface.

**Actual behavior**
The VPN tunnel overwrites the settings of the existing interface in /etc/config/network, causing it to stop functioning.

Components
NethSecurity version: 8.3

Tbaile:
Image: waiting for https://github.com/NethServer/nethsecurity-ui/pull/425

QA:
- Install Nethsecurity with the provided image
- Try to create a server VPN tunnel with the same name of any of the interfaces set up in the firewall
- Ensure validation issues pop up

mamengoni:
Tested and verified:

![Image](https://github.com/user-attachments/assets/dc5d2d46-aea5-48d5-9df7-2184e192ba0b)

PR #883, opened 2024-10-31
ns-api(openwrt): avoiding overwrite of network config

Tbaile:
Ref:
- https://github.com/NethServer/nethsecurity/issues/879


gsanchietti (changes requested):
While the code looks good, it's better to add the same check also on add_client function

gsanchietti (approved):

Issue #912, opened 2024-11-13
Firewall objects: host set modifications not reflected in nft rules


gsanchietti:
**Steps to reproduce**
- Access the host set page in the firewall settings.
- Modify an existing host set by either adding or removing an IP.
- Apply the changes.

**Expected behavior**
- The changes should be reflected in nft rules automatically when an IP is added or removed.

**Actual behavior**
- Adding an IP updates the rules as expected. However, removing an IP does not reflect changes in nft rules.

Executing `fw4 print` shows the correct ipset, but nft rules are not updated accordingly.

**Workaround**
Execute `fw4 restart` to apply all changes.

**Components**
NethSecurity version: 8-23.05.5-ns.1.3.0

**References**
See also #894

Tbaile:
QA:
- install the following image [23.05.5-ns.1.3.0-36-g2f28b1b](https://github.com/NethServer/nethsecurity/actions/runs/11819633844/artifacts/2182453410)
- reproduce the issues above and check that `nft` tables are correctly regenerated with the updated ips

francio87:
Confirm Fixed ✅ 

```sh
root@NethSec:~# nft list set inet fw4 ns_4f8daa46_ipset
table inet fw4 {
        set ns_4f8daa46_ipset {
                type ipv4_addr
                flags interval
                auto-merge
                elements = { 8.8.8.8, 9.9.9.9,
                             80.17.99.72/29 }
        }
}
```

**Removed IP `9.9.9.9` from the object:**

```sh
root@NethSec:~# uci changes
-objects.ns_53667782.ipaddr
objects.ns_53667782.ipaddr+='80.17.99.72/29'
objects.ns_53667782.ipaddr+='8.8.8.8'
```

**The IP set has been correctly updated:**

```sh
root@NethSec:~# nft list set inet fw4 ns_4f8daa46_ipset
table inet fw4 {
        set ns_4f8daa46_ipset {
                type ipv4_addr
                flags interval
                auto-merge
                elements = { 8.8.8.8, 80.17.99.72/29 }
        }
}
```

PR #914, opened 2024-11-13
fix(ns-api): restarting fw4 when updating an ipset

Tbaile:
Ref:
- #912

### DIFF
--- a/packages/ns-api/files/post-commit/reload-ipsets.py
+++ b/packages/ns-api/files/post-commit/reload-ipsets.py
@@ -17,5 +17,10 @@ if 'firewall' in changes:
             force_reload = True
             break
 
+if 'objects' in changes:
+    force_reload = True
+
 if force_reload:
     subprocess.run(["/sbin/fw4", "restart"])
+    # reload dns objects, due to restart they're emptied out
+    subprocess.run(["/usr/bin/ns-objects-reload-dns"])


packages/ns-api/files/post-commit/reload-ipsets.py
```diff
@@ -17,5 +17,11 @@
             force_reload = True
             break
 
+if 'objects' in changes:
+    for change in changes['objects']:
+        if 'ipaddr' in change:
+            force_reload = True
+            break
```
gsanchietti, line 24: Let's force reload at any object change and call the script that reload all ipsets after fw4 restart:
```suggestion
if 'objects' in changes:
    force_reload = True
```

packages/ns-api/files/post-commit/reload-ipsets.py
```diff
@@ -17,5 +17,11 @@
...
+
 if force_reload:
     subprocess.run(["/sbin/fw4", "restart"])
```
gsanchietti, line 27:
```suggestion
    subprocess.run(["/sbin/fw4", "restart"])
    # make sure all domain-set ipset are updates
    subprocess.run( ... )
```

packages/ns-api/files/post-commit/reload-ipsets.py
```diff
@@ -17,5 +17,11 @@
             force_reload = True
             break
 
+if 'objects' in changes:
+    force_reload = True
+
+
```
gsanchietti, line 23:
```suggestion
```

gsanchietti (approved):

Issue #890, opened 2024-11-07
Real time monitoring: more info about connectivities


andre8244:
Inside the connectivity section, the list of active WANs should include also the following info:
- Public IP
- QoS status and assigned bandwidths (if enabled)

andre8244:
**Testing version**

[23.05.5-ns.1.3.0-34-g7f6d83d](https://github.com/NethServer/nethsecurity/actions/runs/11818930546/artifacts/2182240831)

**Test case 1**

- Go to **Network > Interfaces and devices** and configure a single WAN interface
- Go to **Monitoring > Real time monitor > Connectivity**
- Verify that WAN IP address is shown inside **WANs** table
- Go to **Network > Interfaces and devices** and apply a different configuration to WAN interface (e.g. IPv4 vs IPv6, DHCP vs Static address)
- Go to **Monitoring > Real time monitor > Connectivity**
- Verify that WAN IP address is shown inside **WANs** table


**Test case 2**

- Go to **Network > Interfaces and devices** and configure multiwan
- Go to **Monitoring > Real time monitor > Connectivity**
- Verify that the correct IP address of every wan configured is shown inside **WANs** table


**Test case 3**

- Configure one or more WAN interface inside **Network > Interfaces and devices**
- Go to **Network > QoS**
- Enable quality of service for a WAN interface
- Go to **Monitoring > Real time monitor > Connectivity**
- Verify that QoS bandwiths are shown inside **WANs** table

andre8244:
Updated testing image: [23.05.5-ns.1.3.0-49-g597f6078](https://github.com/NethServer/nethsecurity/actions/runs/11937235707/artifacts/2214097449)

PR #923, opened 2024-11-19
feat(ns-api): add ns.report get-public-ip-addresses

andre8244:
Add `ns.report` `get-public-ip-addresses` API

Ref:
- https://github.com/NethServer/nethsecurity/issues/890


gsanchietti (approved):
Ok, only if the util function does not raise an exception: otherwise the API will fail

Issue #937, opened 2024-11-26
Edit `Traffic` section of `Realtime monitoring`


Tbaile:
Due to removal of info by #936 in the `Traffic` section, new tables must be added under every graph.

The figma reference [is here](https://www.figma.com/design/7pL49QCOS2o6fP2JboLhvl/NethSecurity-design?node-id=16819-8768&t=zibtgC6bfmOTkiuc-4).

Issue #900, opened 2024-11-12
Realtime monitoring: traffic details


gsanchietti:
### **Extension Card: Detailed Traffic per Host**

This card extends the detailed traffic monitoring functionality by focusing on local and remote hosts, as well as app-specific traffic.

### **1. App Selection: Traffic per Local Host Using an App**
- **Task:** Select an app to view a ranking of local hosts using the app and their traffic.

### **2. Local Host Selection: App & Remote Host Breakdown**
- **Task:** Select a local host to view:
  - Apps used by the host (like ntopng data).
  - Remote hosts contacted by the host (like ntopng data).
  - Display app and remote host info together to show which host is contacted and via what type of traffic.

### **3. Remote Host Selection: Traffic Breakdown**
- **Task:** Select a remote host to view:
  - Local hosts that contacted it (ranking).
  - Apps or protocols used to connect to that remote host.

Tbaile:
Going to split this down in multiple issues, just to be aware of the tasks that are needed to be done (and to make life easier for reviewers of both UI and backend).

Tbaile:
Image: [23.05.5-ns.1.3.0-89-g31e296d1](https://github.com/NethServer/nethsecurity/actions/runs/12177295129/artifacts/2278349827)

There's not really a QA about this feature but the following are to be browsed to check if the feature is working as intended:

Go to `Monitoring`  > `Real time monitor` and navigate into the `Instant traffic` tab, check that data is visible correctly.

Then you can go back to the `Traffic` tab and try the new filtering options. You can click table records to filter the data by that "client", "application", "remote host" or "protocol.

Suggesting to test the feature with data from a high traffic firewall.

gsanchietti:
Released on 23.05.5-ns.1.4.0

PR #940, opened 2024-11-29
refactor(ns-api): refactored dpi report stats reporting

Tbaile:
Due to changes needed for filtering in https://github.com/NethServer/nethsecurity/issues/937 the code was refactored to offer the functionality needed for the UI feature implementation.

~~To avoid issues, merge only after implementation coming from https://github.com/NethServer/nethsecurity/issues/900.
This can, in fact, be safely added as an additional feature while deprecating the other API to avoid breaking changes.~~

The old `summary` has been backported, this can be safely merged.

### DIFF
--- a/packages/ns-api/files/ns.dpireport
+++ b/packages/ns-api/files/ns.dpireport
@@ -13,6 +13,8 @@ import glob
 import socket
 from datetime import date
 
+BASE_PATH = '/var/run/dpireport'
+
 def reverse_dns(ip):
     try:
         return socket.gethostbyaddr(ip)[0]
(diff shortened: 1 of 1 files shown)


packages/ns-api/files/ns.dpireport
```diff
@@ -26,94 +28,158 @@ def list_days():
...
+def _extract_data(dpi_file: str):
+    with open(dpi_file, 'r') as file:
+        json_file = json.load(file)
+        data = {
+            "total": json_file["total"],
+            "protocol": json_file["protocol"],
+            "application": json_file["application"],
+            "host": json_file["host"]
+        }
+        return data
```
gsanchietti, line 56: The format inside the file is already the one describe, why not just returning it?
```suggestion
        return json.load(file)
```

Also it's better to put a try/except and return an empty object in case of error

Tbaile: Fixed, careful for the next review, cause I've actually backported the `summary` function to avoid breaking changes.

gsanchietti (approved):

Issue #938, opened 2024-11-26
Filtering for traffic


Tbaile:
In the `Traffic` section of the `Realtime monitoring`, should be possible to filter with proper selection of host and one and only one of the following:
- application
- remote host
- protocol

Figma Reference: https://www.figma.com/design/7pL49QCOS2o6fP2JboLhvl/NethSecurity-design?node-id=16819-8768&t=gciUjIwYAUcFe0Ft-4

PR #949, opened 2024-12-03
feat(dpi): adding more filtering options

Tbaile:
Due to needs of https://github.com/NethServer/nethsecurity-ui/pull/452 the filtering section must be expanded adding protocol and remote host filtering.

Ref:
- https://github.com/NethServer/nethsecurity/issues/938


gsanchietti (approved):
:man_shrugging:

PR #990, opened 2024-12-23
fix(ns-api): gracefully handle error when checking for updates

Tbaile:
If for any reason the upstream update server goes offline (both community and enterprise repositories), several issues arise while viewing the Updates page.
By adding the `python3-requests` package it's possible to figure what is going wrong and what happened to the request, leaving the UI with specific error messages that allows to tell the user exactly what is going on.

### DIFF
--- a/packages/ns-api/files/ns.update
+++ b/packages/ns-api/files/ns.update
@@ -11,9 +11,8 @@ import os
 import sys
 import json
 import time
-import urllib
 import subprocess
-import urllib.request
+import requests
 from nethsec import utils
 from urllib.parse import urlparse
 from euci import EUci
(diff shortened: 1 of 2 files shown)


packages/ns-api/files/ns.update
```diff
@@ -89,19 +88,30 @@ def install_package_updates():
 def check_system_update():
     e_uci = EUci()
     current_version = get_system_version()
-    ret = {"currentVersion": f'NethSecurity {current_version}', "lastVersion": "", "scheduledAt": get_update_schedule()}
+    data = {"currentVersion": f'NethSecurity {current_version}', "lastVersion": "", "scheduledAt": get_update_schedule()}
+    url = e_uci.get('ns-plug', 'config', 'repository_url')
```
gsanchietti, line 92: This could raise an exception if repository_url is not set. Please make sure to catch the exception as before.

gsanchietti (changes requested):
Handlers for python-requests call are ok, but restore missing handler for uci.get

PR #992, opened 2025-01-02
mwan3: run user scripts in backgorund

filippocarletti:
/etc/mwan3.user should finish in a timely manner.
See https://openwrt.org/docs/guide-user/network/wan/multiwan/mwan3#alertsnotifications
As of today, the only script is send-mwan-alert which can take 3 minutes.

https://github.com/NethServer/nethsecurity/issues/1014


filippocarletti:
Now that we have forked mwan3, we could avoid using uci-defaults to modify /etc/mwan3.user to ease fixing existing machines: let's talk about it. :-)

gsanchietti (approved):
Please open an issue and link it to the PR

PR #993, opened 2025-01-02
ns-plug: mwan3 hooks needs bash

filippocarletti:
The script uses a bash-only syntax. Using bash gives the following error: /usr/libexec/ns-plug/mwan-hooks: line 17: failed_count++: not found


gsanchietti (approved):

filippocarletti:
Test case:
* check that last line of /etc/mwan3.user ends with '&' for both
  * package update
  * image update
 
In case of package update, please reboot after update and before testing.

PR #994, opened 2025-01-02
fix(ns-plug): avoid mwan3 alerts during restart

filippocarletti:
Running /etc/init.d/mwan3 restart produces a rapid disconnect/connect sequence which triggers an alert lasting a few seconds. This PR avoid the false alarm delaying disconnect alerts by 30 seconds.

https://github.com/NethServer/nethsecurity/pull/992 required.

See #1054

### DIFF
--- a/packages/ns-plug/files/send-mwan-alert
+++ b/packages/ns-plug/files/send-mwan-alert
@@ -9,6 +9,7 @@
 lk=$(uci -q get ns-plug.config.system_id)
 secret=$(uci -q get ns-plug.config.secret)
 url=$(uci -q get ns-plug.config.alerts_url)"alerts/store"
+pidfile="/tmp/mwan3.$INTERFACE"
 
 # Do not send alert if system_id or secret is not set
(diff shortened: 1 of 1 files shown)


filippocarletti:
The only issue remaining is the following log:
```
mwan3-hook: Script /usr/libexec/mwan-hooks/send-mwan-alert failed with exit code 129
```
Exit code 129 means that the process has been killed with SIGHUP (128 + 1).
We could handle 129 as a special case in /usr/libexec/ns-plug/mwan-hooks

gsanchietti (commented):
I really do not like this approach: it feels too hacky

Feel free to merge if your tests were good enough.
Before merge, please open an issue for it

Tbaile:
Talked /w Giacomo, while we think the change is fine, there's the need to open two PRs for this changes, one for the timeout change and the other for the alert during restart, so the work can be tracked easily (and reverted in even easier manner)

Ping me internally if you want me to do it

gsanchietti (approved):

Issue #1010, opened 2025-01-09
IPS Settings Page


Tbaile:
Add a settings page for IPS, this will need additional APIs to be developed.

PR #1011, opened 2025-01-09
feat(snort): adding API for UI

Tbaile:
Changes:
- Added API to read configuration
- Added API to check oinkcode (this is done by downloading the registered rules, not efficient but the only way)
- Removed `ns_disabled_rules` from `setup` API, this will need a separate API.
- Now service restarts at every change in `snort` config
- Added ability to comment bypasses 

~~NOTE: now the service downloads the new rules at every restart. It's not optimal, but:~~
~~1. checking if the files in `rules` exists, is not a indicator of the need the rules again (I.E. if you are a community user and then add the "oinkcode" later), this locks the user to the old rules until the cronjob applies~~
~~2. there's no way to know what changed in the configuration, this was useful in case we want to re-download the rules and reapply them, only if the oinkcode changed.~~

~~If you think that point 1 can be overruled and it's fine, I can revert the [change](https://github.com/NethServer/nethsecurity/pull/1011/files#diff-793a7bc7e42088934e22255570bc234ffbc9b7d5945eb52d5ab46176b36a2e4eR35) no problem.~~

Now we check if the rules need to be downloaded by referring to the downloaded file. Rules are deleted when oinkode is present or not, so that when you add a code, fresh rules are downloaded.

Ref:
- https://github.com/NethServer/nethsecurity/issues/1010

### DIFF
--- a/packages/ns-api/README.md
+++ b/packages/ns-api/README.md
@@ -7792,6 +7792,65 @@ If `include_vpn` is `true`, the API will include the VPN networks in the `HOME_N
 The `ns_policy` can be `balanced`, `security` or `connectivity` or `max-detect`.
 The `ns_disabled_rules` is a list of SIDs (integer) of rules to be disabled.
 
+### status
+
+Returns the configuration set for `snort`.
+
+```bash
+api-cli ns.snort status
+```
+
+Response example:
+
+```json
+{
+  "enabled": true,
+  "ns_policy": "connectivity",
+  "oinkcode": "123456789"
+}
+```
+
+### save
+
+Set `snort` configuration
+
+```bash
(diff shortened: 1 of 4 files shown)


filippocarletti:
Quoting from https://www.snort.org/snort_license
> Cisco reserves the right to limit the time and/or frequency that the Rules are made available for download at [www.snort.org](https://www.snort.org/)

Download may fail if we get rate-limited and/or blacklisted.

Tbaile:
Updated the behaviour, this now:
- If rules were downloaded, do not download them again (wait for the night script to apply updates)
- When oinkcode is set (or not) delete the other set of rules, this makes that fresh rules are downloaded only when it's being switched from community to registered rules.

PR #1021, opened 2025-01-14
fix(core): disable failsafe

filippocarletti:


Tbaile (approved):
If it builds and works :shrug:

gsanchietti (commented):
Please fix:
- add a comment for the issue description, better if a link to an issue
- the commit message does not respect the convention rules

Tbaile:
Closing due to being impossible to predict the outcome.
This might cause a heavy burden on the support team, since this change makes impossible to reset or recover the root password of the system.
Let's find another way.

Issue #1018, opened 2025-01-13
snort: Suppressed alerts page


Tbaile:
Refs:
- backend PR: https://github.com/NethServer/nethsecurity/pull/1033
- frontend PR: https://github.com/NethServer/nethsecurity-ui/pull/487

PR #1033, opened 2025-01-16
feat(ips): adding suppression alerts

Tbaile:
Ref:
- https://github.com/NethServer/nethsecurity/issues/1018

### DIFF
--- a/packages/ns-api/files/ns.snort
+++ b/packages/ns-api/files/ns.snort
@@ -353,6 +353,68 @@ def __enable_rule():
     e_uci.save('snort')
 
 
+def __list_suppressed_alerts():
+    e_uci = EUci()
+    suppressed_alerts = []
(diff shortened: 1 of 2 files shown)


packages/ns-api/files/ns.snort
```diff
@@ -353,6 +353,78 @@ def __enable_rule():
...
+        valid_ip = True
+    except:
+        pass
+    try:
+        ipaddress.IPv4Network(request['ip'])
+        valid_ip = True
+    except:
+        pass
+    if not valid_ip:
+        raise ValidationError('ip', 'invalid')
```
gsanchietti, line 399: Something like this should work:
```suggestion
    try:
        ipaddress.ip_network(request['ip'])
    except:
        raise ValidationError('ip', 'invalid')
```

Tbaile: It does, applying now

gsanchietti (approved):

Issue #906, opened 2024-11-12
UI for Threat Shield DNS


gsanchietti:
**Description:**
Create user interface for managing Threat Shield DNS. Currently, Threat Shield DNS is primarily configured via the command line, as detailed in the [documentation](https://docs.nethsecurity.org/en/latest/threat_shield_dns.html#threat-shield-dns-section). A UI would enhance user experience by providing an easier way to manage DNS-based blocking and other functionalities.

**Proposed solution:**
- Page to view and manage DNS blocklists.
- Page to load community-maintained blocklists or enterprise feeds and select one or more list.
- Page to create bypass for certain hosts on the local network.
- Logs and reports for DNS requests and blocked domains, if possible.

Tbaile:
Before closing the issue, remember the documentation!

gsanchietti:
Overall feature works well, but I've found 2 issues: not verified.

**First issue**

Local blocklist does not work: check `/tmp/dnsmasq.d/adb_list.overall` file
To make it work:
```
/etc/init.d/adblock restart
```
After the restart, the blocklist contains also an invalid entry for `example.com`

**Second issue**

The option "Force DNS redirection on these zones" should not list WAN zone.

gsanchietti:
We can ignore the presence of `example.com` domain.

From adblock doc:
```
adb_lookupdomain	example.com	external domain to check for a successful DNS backend restart or 'false' to disable this check
```

From dnsmasq doc:
```
The special server address '#' means, "use the standard servers", so --server=/google.com/1.2.3.4 --server=/www.google.com/# will send queries for google.com and its subdomains to 1.2.3.4, except www.google.com (and its subdomains) which will be forwarded as usual.
```

andre8244:
### Updated testing image

Artifact of https://github.com/NethServer/nethsecurity/actions/runs/13057932549

gsanchietti:
Verified

PR #1040, opened 2025-01-21
fix(ns-api): fix threat shield dns apis

andre8244:
Ref:
- https://github.com/NethServer/nethsecurity/issues/906

### DIFF
--- a/packages/ns-api/files/ns.threatshield
+++ b/packages/ns-api/files/ns.threatshield
@@ -380,10 +380,7 @@ def dns_edit_blocklist(e_uci, payload):
     return {'message': 'success'}
 
 def dns_list_settings(e_uci):
-    try:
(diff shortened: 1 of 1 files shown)


packages/ns-api/files/ns.threatshield
```diff
@@ -380,15 +380,19 @@ def dns_edit_blocklist(e_uci, payload):
     return {'message': 'success'}
 
 def dns_list_settings(e_uci):
+    try:
+        ts_enabled = e_uci.get('adblock', 'global', 'ts_enabled')
+    except:
+        ts_enabled = '0'
```
gsanchietti, line 386: This could be simplified:
```suggestion
     ts_enabled = e_uci.get('adblock', 'global', 'ts_enabled', default='0')
```

gsanchietti (approved):

Issue #989, opened 2024-12-20
Allow Forward Rules Between the Same Zone or Any Zone


cotosso:
**Description:**
In the current implementation of the forward rules, it is not possible to create rules where the source and destination are within the same zone, a specific zone, or the "any" option. This limitation results in several practical issues and inefficiencies.

**Examples:**

- Rules Between Two LAN Zones:

It is currently impossible to create a rule using two LAN zones. For instance, if I have multiple LAN interfaces (each with its own CIDR), I am forced to specify one zone as LAN and the other as ANY.

- Rules Using "Any" for Both Source and Destination:

If I want to block all traffic between zones and allow specific traffic only on certain ports using host sets, I am required to create multiple rules instead of a single, comprehensive rule.
On NethSecurity 8, the "any" option operates only within the forward context. As a result, it cannot block access to the firewall itself. 

**Proposed Solution:**
Relax form validator to:
* Allow the creation of forward rules where the source and destination are the same zone.
* Enable the use of the "any" option for both source and destination in forward rules.

andre8244:
**Testing image**

[23.05.5-ns.1.4.1-37-g5ce686044](https://github.com/NethServer/nethsecurity/actions/runs/12986395394/artifacts/2489596352)

**QA**

- Check that it's possible to create a forward rule with the same source/destination zone
- Check that it's possible to create a forward rule selecting **Any** as source and destination zone

gsanchietti:
Verified: it's now possibile to create such rules that are correctly written inside nft.

PR #1042, opened 2025-01-23
fix(firewall): allow rules with same src/dest zone

andre8244:
Ref:
- https://github.com/NethServer/nethsecurity/issues/989


Tbaile:
Release the library and the UI, add the bumps here so that's not only a docs PR

PR #1052, opened 2025-01-29
fix!: NIS2 Adjustments

Tbaile:
PR that sums all the changes needed to strenghten up the firewall security-wise.

This will be on hold since 8.6.


Tbaile:
Changes already incorporated by https://github.com/NethServer/nethsecurity/issues/1164

PR #1080, opened 2025-02-13
wip: trying to add board files

Tbaile:


cotosso:
We also need to add Z9/Z9+ board file, cause if an SFP expansion is present upgrading from 23.05.05 to 24.10 could move SFP ports from eth0/1/2/3 to eth8/9/10/11

Issue #1028, opened 2025-01-15
Disable SNMP by default


cotosso:
SNMP is enabled by default on the system, potentially exposing sensitive information such as interfaces, and IP addresses, which could be exploited in an attack.

**Proposed Solution**

Disable SNMP by default in the system configuration.
Provide an option to enable SNMP only if explicitly required by the user, ensuring it is not active in standard deployments.

Tbaile:
QA:

- Update a 1.4.1, check that the service is still running `snmpwalk -v 2c -c public 127.0.0.1`
- Fresh install, the command `snmpwalk -v 2c -c public 127.0.0.1` shouldn't return anything

NOTE: command must be ran inside the unit.

cotosso:
1. Update a 1.4.1, check that the service is still running snmpwalk -v 2c -c public 127.0.0.1 : **VERIFIED**

2. Fresh install, the command snmpwalk -v 2c -c public 127.0.0.1 shouldn't return anything : **VERIFIED**

![Image](https://github.com/user-attachments/assets/36bd815c-1642-427a-8876-efbbe7c7a92a)

PR #1091, opened 2025-02-20
fix(snmpd): disable config by default

Tbaile:
Ref:
- https://github.com/NethServer/nethsecurity/issues/1028


gsanchietti (approved):
Before merge, please fix also the doc: https://docs.nethsecurity.org/en/latest/snmp.html

Tbaile:
Shall we add a note that from fresh installed NethSecurity instances from v1.5.0 this is disabled by default while updates from 1.4.1 have it enabled?

Issue #1126, opened 2025-03-14
Multiwan: firewall persistently routes via lowest metric WAN, even if offline


francio87:
The traffic generated directly by the firewall does not follow the rules defined in the MultiWAN section. As a result, if the WAN interface with the lowest metric loses connectivity, the firewall itself can no longer access the internet, even if other WAN connections are still available.

**Steps to reproduce**

- Configure a MultiWAN setup with at least two WAN connections
- Simulate an internet access failure on the WAN with the lowest metric (disconnecting the cable won't show the issue)

**Expected behavior**

- The firewall switches automatically to the next available WAN connection for its own traffic
- Internet access, DNS resolution, and VPN connectivity remain operational
- Clients using the firewall as DNS resolver can still resolve names

**Actual behavior**

- The firewall always attempts to route its own traffic through the WAN with the lowest metric, even if that connectivity is offline
- VPN tunnel client connections also cannot be established after the WAN with the lowest metric loses access
- Internet connectivity for the firewall fails, and DNS resolution for clients using the firewall as a DNS server fails too, while other client traffic is correctly routed according to the MultiWAN rules and policies.

Tbaile:
Image: [24.10.0-ns.1.4.99-alpha1-35-g010b308](https://github.com/NethServer/nethsecurity/actions/runs/13964479135/artifacts/2786482664)

Check issue is not reproducible by checking with `ip ru`.

PR #1133, opened 2025-03-18
fix(mwan3): fix wrong traffic routing with specific metrics

Tbaile:
By leveraging /etc/mwan3.user, we can generate ip rules that force traffic to be routed to specific interfaces. This PR reverts 05fbb03d1d1fae30f2fe486dcffde3772c2d76f4 and implements back the functionality.

Ref:
- https://github.com/NethServer/nethsecurity/issues/1126

### DIFF
--- a/packages/mwan3/Makefile
+++ b/packages/mwan3/Makefile
@@ -102,6 +102,8 @@ define Package/mwan3/install
 		$(1)/usr/sbin/
 	$(INSTALL_BIN) ./files/usr/sbin/mwan3track \
 		$(1)/usr/sbin/
+	$(INSTALL_BIN) ./files/usr/sbin/mwan3routertraffic \
+		$(1)/usr/sbin/
 
 	$(INSTALL_DIR) $(1)/etc
 	$(INSTALL_BIN) ./files/etc/mwan3.user \
@@ -111,6 +113,8 @@ define Package/mwan3/install
 
 	$(INSTALL_DIR) $(1)/etc/uci-defaults
 	$(INSTALL_DATA) ./files/etc/uci-defaults/mwan3-migrate-flush_conntrack \
+		$(1)/etc/uci-defaults/
+	$(INSTALL_DATA) ./files/etc/uci-defaults/mwan3-addroutertraffic-execution \
 		$(1)/etc/uci-defaults/
 endef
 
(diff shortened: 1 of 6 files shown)


packages/mwan3/files/usr/sbin/mwan3routertraffic
```diff
@@ -0,0 +1,55 @@
...
+# check for passed parameters, if empty use environment variables
+interface_name="${1:-$INTERFACE}"
+action_name="${2:-$ACTION}"
+
+config_load mwan3
+
+# Get the id of the interface handled by mwan3, if empty, exit
+id=""
+mwan3_get_iface_id id "$interface_name"
+[ -n "$id" ] || return 0
```
gsanchietti, line 26: Use exit instead of return

packages/mwan3/files/usr/sbin/mwan3routertraffic
```diff
@@ -0,0 +1,55 @@
...
+config_get family "$interface_name" family ipv4
+# exit if the interface is not supported
+if [ "$family" = "ipv4" ]; then
+  IP="$IP4"
+  network_get_ipaddr wan_addr "${1}"
+elif [ "$family" = "ipv6" ] && [ $NO_IPV6 -eq 0 ]; then
+  IP="$IP6"
+  network_get_ipaddr6 wan_addr "${1}"
+else
+  return 1
```
gsanchietti, line 40: Use exit instead of return

packages/mwan3/files/usr/sbin/mwan3routertraffic
```diff
@@ -0,0 +1,55 @@
...
+  $IP rule add pref $((id+1500)) from "$wan_addr" lookup "$id"
+elif [ "$action_name" = "disconnected" ]
+then
+  # Remove the rule from the routing table
+  $IP rule del pref $((id+1500))
+fi
+
+
+
+
```
gsanchietti, line 55: Cleanup empty lines

packages/mwan3/Makefile
```diff
@@ -8,9 +8,8 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mwan3
-# version 2.11.17 is from Nethesis: it clears the mwan3 config
```
gsanchietti, line 11: Do no remove this line.
Change like:
```
from version 2.11.17 is Nethesis fork: it clears the mwan3 config
```

Tbaile:
Still not sure to move the files in `/usr/libexec/mwan-hooks/`, I'll give it some thoughts.

gsanchietti (approved):
This looks good, but I propose a couple of changes:
1. just drop mwan3routertraffic inside ns-plug under the mwan scripts directory
2. if 1 is not good for you, at least move /usr/sbin/mwan3routertraffic under libexec directory because it's not a script that should be called manually

Issue #1135, opened 2025-03-20
Migration: netifyd does not get configured


Tbaile:
**Steps to reproduce**

- migrate any ns8 installation

**Expected behavior**

- netifyd correctly filters traffic and makes stats

**Actual behavior**

- still uses the "default installation" interfaces, breaking analisis and more

**Components**

ns-import 0.0.19

Tbaile:
Image: [24.10.0-ns.1.4.99-alpha1-37-g6761540](https://github.com/NethServer/nethsecurity/actions/runs/14029828920/artifacts/2805766488)

cotosso:
Verified:
after the migration netifyd has the right configuration and it works flawlessy.


![Image](https://github.com/user-attachments/assets/21b68420-a8e4-46eb-9a75-05161ca05132)

PR #1138, opened 2025-03-24
fix(ns-migration): configure netifyd on migration

Tbaile:
Ref:
- https://github.com/NethServer/nethsecurity/issues/1135


gsanchietti (approved):
Very nice solution!

Tbaile:
Going to test this without merging, since two PRs are editing the same file.

https://github.com/NethServer/nethsecurity/pull/1139

Issue #1136, opened 2025-03-20
Migration: default logging rules are not being applied


Tbaile:
**Steps to reproduce**

- migrate a firewall installation with logging enabled in rules or zones

**Expected behavior**

- logging is enable and limited by defaults introduced by https://github.com/NethServer/nethsecurity/issues/1105

**Actual behavior**

- logging is unlimited, probably causing issues when high traffic happens

**Components**

ns-import 0.0.19

Tbaile:
This QA is being done without merging the PR associated.

Image: [24.10.0-ns.1.4.99-alpha1-37-ga0f84e4](https://github.com/NethServer/nethsecurity/actions/runs/14029849848/artifacts/2805800349)

cotosso:
Verified:
Migrated rules are when logged and are subject to rate limiting as well as the new ones.

![Image](https://github.com/user-attachments/assets/e03b1348-9765-4e06-9617-71fd99dd2bb7)

![Image](https://github.com/user-attachments/assets/ad9bb04b-5e7b-4552-9313-5fac66213902)

![Image](https://github.com/user-attachments/assets/4d7209d8-3249-4405-aa64-52cd54e54cce)

PR #1139, opened 2025-03-24
fix(ns-migration): apply default logging when migrating

Tbaile:
Ref: 
- https://github.com/NethServer/nethsecurity/issues/1136


gsanchietti (approved):

Tbaile:
Going to test this without merging, since two PRs are editing the same file.

https://github.com/NethServer/nethsecurity/pull/1138

PR #1142, opened 2025-03-25
fix(announce_updates): exclude PRs

gsanchietti:
Previously the script was generating an empty announce during days when only pull requests were closed.


Tbaile (approved):
:shrug:

PR #1147, opened 2025-03-27
perf(ns-ui): halving load time for UI using gzip

Tbaile:
Using gzip in nginx, we halve the load time for the ui, this applies to good and bad connection situations.


gsanchietti (approved):
Looks good to me, just pay attention to the indentation

Issue #1146, opened 2025-03-27
Backup uploaded/changed every day


filippocarletti:
The tar.gz backup file changes every time the backup is run.
Two causes:
1. timestamp of /etc/backup/installed_packages.txt
2. /etc/uci-defaults/10_disable_services

**Steps to reproduce**

- Run `sysupgrade -k -b /tmp/backup1.tar.gz`
- Run `sysupgrade -k -b /tmp/backup2.tar.gz`
- Compare the two backups: `md5sum  /tmp/backup?.tar.gz`

**Expected behavior**

The backups should be identical (the content is).

**Actual behavior**

Backups are different.

**See also**

https://github.com/openwrt/openwrt/pull/16146/files
https://github.com/openwrt/openwrt/issues/16145

gsanchietti:
Proposed implementation in #1149 can cause many regressions.

As @filippocarletti proposal, let's wait a fix from upstream.

gsanchietti:
The  `/etc/backup/installed_packages.txt` is rewritten every time so the modification date of the file changes at every backup.
Keeping the modification date consistent across backup is not a a feature pursued by upstream implementation.

During the analysis, I've found also another case that will change the backup: every time adblock is restated, it generates the local allowlist and blocklist.

On machines with a subscription, the backup is updated every night to the remote server and this is just fine for normal usage.
In case a configuration snapshot is required, the user can still download the backup or push it manually to the remote using the web UI.

I'm closing the issue as CANTFIX.
Feel free to reopen or create a new if a fix for all cases is available.

Tbaile:
Reopening issue due to @filippocarletti wanted to explore a new approach into checking if the backup is the same as the old one. Opened related PR with a concept I did a while ago.

PR #1149, opened 2025-04-01
fix(send-backup): change compare logic

gsanchietti:
The backup contains files that are changing the date but not the content: generated tar gz always changes the md5 making it impossibile to know if the backup must be sent to the remote server.

Changes:
- new compare logic: use the md5 of each file instead of the md5 of the backup file
- do not use sysupgrade to create the file: it internally creates the `/etc/backup/installed_packages.txt` file so there is not way to compare the content using md5
- add exclusion support using `/etc/backup.exclude`: add here the files that need to be excluded
- added logging to messages and to stderr (only if the script has been called from a tty)

The script logs only when sending the backup. Example of log:
```
Apr  2 07:39:32 NethSec send-backup: [INFO] Backup upload completed successfully
```

If no password is set, also a warning will be logged:
```
Apr  2 07:39:29 NethSec send-backup: [WARNING] No passphrase found, sending unencrypted backup. This method is deprecated and will be removed in the future.
```

Note: the archive generated using `sysupgrade -k -b` contains an extra file named `/etc/uci-defaults/10_disable_services` which is not needed in our installations.

Reference: #1146

### DIFF
--- a/packages/ns-plug/Makefile
+++ b/packages/ns-plug/Makefile
@@ -31,6 +31,7 @@ endef
 
 define Package/ns-plug/conffiles
 /etc/config/ns-plug
+/etc/backup.exclude
 endef
 
 # this is required, otherwise compile will fail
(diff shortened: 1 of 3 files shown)


Tbaile (approved):
Seems ok, I'm feeling a lot of unexplainable regressions coming from this one

gsanchietti:
I agree with you: it's too risky.
I'm closing it and add a note.

PR #1150, opened 2025-04-02
fix(ns-api): mark templates file as non-config (again)

gsanchietti:
Changes on commit 3049dabc52d6f209b9a82e2e098e545ebcfc0e9f were not enough.


gsanchietti:
Avoid this change for now: if modifications are needed for the templates db, we are going to use uci defaults scripts.

PR #1153, opened 2025-04-04
chore(ns-api): version bump

Tbaile:


Tbaile:
Thinking about releasing this in the new 1.5 directly, otherwise many will feel like the new version "is broken"

Issue #924, opened 2024-11-20
Threat shield IP UI: manage list of blocked IPs


gsanchietti:
Create UI and API to manage IPs blocked by Brute force attack feature.

**Dynamic blocklist**
- **View all IP addresses currently on the blocklist:** ~~`/etc/init.d/banip survey blocklistv4`~~
  impossible to parse, maybe solution in `nft -j list set inet banIP blocklistv4 | jq '.nftables[1].set.elem'`
- **Unban an IP address:** `nft delete element inet banIP blocklistv4 { IP_ADDRESS }`

Note: specify the correct blocklist in commands when prompted (blocklistv4 for IPv4, blocklistv6 for IPv6).

**All blocklists**
- **Look up a specific IP in the blocklist:** `/etc/init.d/banip search IP_ADDRESS`

**Tasks:**
1. Create a UI to display all IP addresses currently on the blocklist.
2. Implement a search functionality in the UI to look up a specific IP in the blocklist.
3. Provide an option in the UI to unban an IP address.
4. Develop API endpoints for viewing, searching, and unbanning IP addresses.

Tbaile:
Image: [24.10.0-ns.1.5.1-40-g81ad437f](https://github.com/NethServer/nethsecurity/actions/runs/15165003847/artifacts/3169500475)

gsanchietti:
Verified.
But the UX to search the blocked IP addresses is not very clear. See screenshot:

![Image](https://github.com/user-attachments/assets/f8f04a17-4e96-4e9c-95a3-be8de1ffc5b7)

The search works only on enabled lists: I think it worth mention it somewhere (a tooltip or a description).

PR #1190, opened 2025-04-29
feat(threatshield): active block list and ip search

Tbaile:
Added APIs for feature linked below

Ref: https://github.com/NethServer/nethsecurity/issues/924

### DIFF
--- a/packages/ns-api/files/ns.threatshield
+++ b/packages/ns-api/files/ns.threatshield
@@ -12,6 +12,7 @@ import json
 import urllib
 import urllib.request
 import urllib.error
+import logging
 import ipaddress
 from euci import EUci
 from nethsec.utils import ValidationError, validation_error, generic_error
@@ -563,7 +564,8 @@ def __parse_expression(expression):
                 "val": __parse_expression(expression["elem"]["val"]),
             }
 
-    raise NotImplementedError('Unsupported expression type: {}'.format(type(expression)))
+    logging.warning('Unsupported expression type: {}'.format(type(expression)))
+    return None
 
 
 def __parse_set(blocklist):
@@ -575,7 +577,7 @@ def __parse_set(blocklist):
             and len(blocklist["nftables"]) > 1
             and "set" in blocklist["nftables"][1]
             and "elem" in blocklist["nftables"][1]["set"]):
-        blocked_ips = __parse_expression(blocklist["nftables"][1]["set"]["elem"])
+        blocked_ips = [ip for ip in __parse_expression(blocklist["nftables"][1]["set"]["elem"]) if ip is not None]
 
     return blocked_ips
 
@@ -625,9 +627,7 @@ def list_active_blocks():
 
 def remove_active_block():
     data = json.loads(sys.stdin.read())
-    if "ip" not in data:
-        return validation_error("ip", "required")
-    if data["ip"] == "":
+    if data.get("ip", "") == "":
         return validation_error("ip", "required")
 
     try:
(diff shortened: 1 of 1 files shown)


packages/ns-api/files/ns.threatshield
```diff
@@ -541,6 +542,141 @@ def dns_delete_bypass(e_uci, payload):
...
+
+    return response
+
+
+def remove_active_block():
+    data = json.loads(sys.stdin.read())
+    if "ip" not in data:
+        return validation_error("ip", "required")
+    if data["ip"] == "":
+        return validation_error("ip", "required")
```
gsanchietti, line 631: This could be simplified
```suggestion
    if data.get("ip", "") == "":
        return validation_error("ip", "required")
```

packages/ns-api/files/ns.threatshield
```diff
@@ -541,6 +542,141 @@ def dns_delete_bypass(e_uci, payload):
...
+
+    try:
+        ip = ipaddress.ip_address(data["ip"])
+    except ValueError:
+        return validation_error("ip", "invalid", data["ip"])
+
+    found = False
+    blocks = filter(lambda x: x["ip"] == str(ip), list_active_blocks())
+    for _ in blocks:
+        found = True
```
gsanchietti, line 641: This is very hard to read.
Can we use something like `found = len(blocks) > 0`?

packages/ns-api/files/ns.threatshield
```diff
@@ -541,6 +542,141 @@ def dns_delete_bypass(e_uci, payload):
...
+                __parse_expression(expression["range"][1])
+            ))
+        elif "elem" in expression:
+            return {
+                "timeout": expression["elem"]["timeout"],
+                "expires": expression["elem"]["expires"],
+                "val": __parse_expression(expression["elem"]["val"]),
+            }
+
+    raise NotImplementedError('Unsupported expression type: {}'.format(type(expression)))
```
gsanchietti, line 566: Maybe returning an empty object is safer?

Tbaile: It depends, if we return an empty object we lose the information that something is there, but we have no clue what it is. This way we can catch instantly the issue.

Feel like either road is fine, what do you prefer?

packages/ns-api/files/ns.threatshield
Tbaile: This appears to be the only way, bit ugly but still more readable

```suggestion
    found = len(list(blocks)) > 0
```

packages/ns-api/files/ns.threatshield
gsanchietti: As far as it's handled well by the UI, we can leave as is.

packages/ns-api/files/ns.threatshield
gsanchietti: If possible, I prefer you suggestion which uses len

packages/ns-api/files/ns.threatshield
Tbaile: Oh well, then it doesn't :grimacing: 

I'll return a empty object and add some logging.

packages/ns-api/files/ns.threatshield
```diff
@@ -541,6 +543,138 @@ def dns_delete_bypass(e_uci, payload):
...
+        if len(matches) > 0:
+            return {
+                'found': True,
+                'list': matches[0]
+            }
+    except subprocess.CalledProcessError:
+        return validation_error("banip", "processing")
+
+    return {
+        'found': False
```
gsanchietti, line 674: Do not know if it's better for the UI, or not, but I'd add always add the `list` field to have a uniform return value
```suggestion
        'found': False
        'list': []
```

gsanchietti (approved):

packages/ns-api/files/ns.threatshield
Tbaile: UI takes care of that, without issues, even with typing support: https://github.com/NethServer/nethsecurity-ui/pull/538/files#diff-cf73b5e916d8da4cf7120ea74aefe5eaf0db51e79c3a0ec0fc2a1b7cf03247a7R21

PR #1194, opened 2025-05-03
feat(conf): enable persistent /var

filippocarletti:
Keep DHCP leases across reboot
https://github.com/openwrt/openwrt/commit/57807f50ded6cf0996284a850084183af13d5894


filippocarletti:
This feature needs to be evaluated carefully, postponing and closing.
/var is mostly used by:
* log
* netdata
* snort
* banip
* adblock
* OpenVPN

dpireport stores data in /var/run which is in ram.

Issue #1082, opened 2025-02-17
Ensure Default Log Storage on Persistent Storage Instead of RAM


cotosso:
**Description**
Currently, log files are stored in RAM by default. However, for installations with advanced security requirements, logs should be persistently saved by default.

**Proposed Solution**
To align with security-by-design and security-by-default principles, logging should be enabled immediately upon deployment. A minimally invasive approach that maintains the current behavior while improving security is proposed:

On first UI access, display a modal asking whether to save logs on persistent storage.

The modal must be acknowledged before it disappears.

Until explicitly answered, the modal will reappear on every UI access (e.g., reopening a closed tab).

**Available Choices**

- Log on installation storage partition (default)
- No logging

For cases where an additional storage partition is available, the user can later navigate to the storage settings, disable logging on the current partition, and enable it on the new storage.

**Impact & Benefits**

Ensures critical logs are not lost due to RAM storage volatility.
Provides users with an informed choice while maintaining security defaults.
Allows for flexible log storage reconfiguration post-installation.

**In case of presence of a controller**

In case of use of a controller the admin can link NethSecurity to the controller too, saving logs in both local and remote resources.
This also guarantee log save in case of connectivity issues (including attack scenarios) that may prevent logs from reaching the controller, resulting in loss of critical data.

filippocarletti:
Couldn't we disable the option to have logs in RAM?

cotosso:
> Couldn't we disable the option to have logs in RAM?

yes, it depends from the implementation we are going to use to write all of them in the storage.
We need to make sure that the logs are maintained even after version updates.
If this feature is preserved we can also decide to always write them to the storage.

Tbaile:
Proposed solution: 
Move the logs inside the storage by default (this can, however, be lost during updates) and then if the [Secure Installation Flag](https://github.com/NethServer/nethsecurity/issues/1088) is enabled, prompt the user perpetually to enable the storage logs. This will be forced even if the controller logging is enabled.

Tbaile:
Image: [24.10.0-ns.1.5.1-28-ga635366e](https://github.com/NethServer/nethsecurity/actions/runs/14990611405/artifacts/3112290577)

filippocarletti:
* Test case 1: new ns-install
  * Write the image to a USB key, and use the key to
  * Boot a new hardware, then check with `df` that `/mnt/data` is NOT mounted
  * Run `ns-install`, remove the USB key, and boot the system from permanent storage
[...]

PR #1198, opened 2025-05-05
feat(ns-storage) enable storage by default

filippocarletti:
Enable storage on extra partition on boot device only if not USB.
https://github.com/NethServer/nethsecurity/issues/1082

### DIFF
--- a/files/etc/uci-defaults/29-ns-storage-default
+++ b/files/etc/uci-defaults/29-ns-storage-default
@@ -6,7 +6,7 @@ if readlink "/sys/block/$diskdev" | grep -q "usb"; then
     exit 1
 else
     [ "$(uci -q get fstab.ns_data)" = "mount" ] && exit 0
-    jdev=$(/usr/libexec/ns-storage-setup-partition)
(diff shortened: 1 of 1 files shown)


files/etc/uci-defaults/29-ns-storage-default
```diff
@@ -0,0 +1,13 @@
+. /lib/functions.sh
+. /lib/upgrade/common.sh
+
+export_bootdevice && export_partdevice diskdev 0
+if readlink "/sys/block/$diskdev" | grep -q "usb"; then
+    exit 1
+else
+    [ "$(uci -q get fstab.ns_data)" = "mount" ] && exit 0
+    jdev=$(/usr/libexec/ns-storage-setup-partition)
```
gsanchietti, line 9: If the setup-partition exists with 1, the script should terminate to avoid:
- calling add-storage on an empty variable `$part`
- outputting an error like `Failed to parse json data: unexpected end of data`

gsanchietti:
LGTM

Tbaile (approved):

PR #1199, opened 2025-05-05
feat: disable i915

filippocarletti:
Disable Intel i915 GFX driver to avoid messy logs


Tbaile:
Looks good to me, shall we wait for next image release for the merge?

PR #1210, opened 2025-05-12
build: fix board.d for Z9

filippocarletti:
Handle both Z9 and Z9+

### DIFF
--- a/files/etc/board.d/70nethesis_boxes
+++ b/files/etc/board.d/70nethesis_boxes
@@ -33,21 +33,41 @@ nethesis-nethbox-z11)
   ;;
 nethesis-nethbox-z9)
   if lspci -D | grep -q "SFP+"
-  then
-    ucidef_set_network_device_path "eth0" "pci0000:00/0000:00:01.1/0000:02:00.0" # fiber
-    ucidef_set_network_device_path "eth1" "pci0000:00/0000:00:01.1/0000:02:00.1" # fiber
-    ucidef_set_network_device_path "eth2" "pci0000:00/0000:00:01.1/0000:02:00.2" # fiber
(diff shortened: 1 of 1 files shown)


Tbaile (approved):

Tbaile:
@cotosso pointed out that if the box is a `Z9`, the remapping is not needed, while I am skeptical about it due to the fragility of the PCI order that the vendor supplied, it appears that it sets the ports correctly.

cotosso:
After extensive testing, one key finding is that the default behavior (in the absence of board.json) differs between the Z9 and Z9+ boxes.
The Z9 units we tested consistently place the SFP interfaces **after** the Ethernet ports (so SFP cages are eth8–eth11), both with kernel 5.x (NethSecurity 8.4) and kernel 6.x (NethSecurity 8.5).

The Z9+, on the other hand, consistently places the SFP interfaces **before** the Ethernet ports (so SFP cages are eth0–eth3), also in both version 8.4 and 8.5. As with the Z9, no behavioral change was observed across versions.

To avoid potential issues, it is advisable to stick as closely as possible to the default behavior observed when the system is not managed by a board file. 
If the behavior remains consistent across different versions, **it's not strictly necessary** to enforce interface mappings via board configuration.
By contrast, for other models, it was necessary to rename the interfaces because the upgrade would otherwise reorder them.

The goal, for both hardware models and all software versions mentioned, is to preserve interface ordering consistency with existing installations, so that systems upgraded from version 1.3.0 or 1.4.1 to newer versions do not experience any unexpected changes.

cotosso (approved):
Looks good to me.
It addresses the Z9 issue (where the standard displacement differs from the Z9+)
and prevents unwanted movement on the Z7 when a third-party SFP expansion (not factory-installed) is inserted.

PR #1229, opened 2025-05-23
fix(banip): remove non-existing lists

gsanchietti:
When banip can't download a list, it remains stack for many minutes.
As result, the first configuration seems not working at all from the user perspective.

Changes:
- grab latest releases from upstream: https://github.com/openwrt/packages/commit/7383b4817b2482dab24b15c91311c83cd7fbdea9
- delete non-working lists

### DIFF
--- a/packages/banip/files/banip.feeds
+++ b/packages/banip/files/banip.feeds
@@ -1,59 +1,21 @@
 {
-	"adaway":{
-		"url_4": "https://raw.githubusercontent.com/dibdot/banIP-IP-blocklists/main/adaway-ipv4.txt",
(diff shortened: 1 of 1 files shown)


packages/ns-plug/files/send-inventory
Tbaile, line 1: Something is wrong here!

Tbaile (changes requested):

packages/ns-plug/files/send-inventory
gsanchietti: I was pretty sure to have it fixed :thinking:

PR #1234, opened 2025-05-27
fix(ns-storage): use filesystem label

filippocarletti:
Use ext4 label "ns_data" to identify storage disk/partition

### DIFF
--- a/packages/ns-storage/files/add-storage
+++ b/packages/ns-storage/files/add-storage
@@ -28,10 +28,10 @@ mkdir -p "$mdir"
 mount "$device" "$mdir"
 
 # find uuid and create automount config
-eval $(block info | grep "$device" | awk '{print $2}')
+eval $(block info | grep "$device" | awk '{print $3}')
 uci set fstab.ns_data=mount
 uci set fstab.ns_data.target="$mdir"
-uci set fstab.ns_data.uuid="$UUID"
+uci set fstab.ns_data.label="$LABEL"
 uci set fstab.ns_data.enabled=1
 uci commit fstab
 
--- a/packages/ns-api/files/ns.storage
+++ b/packages/ns-api/files/ns.storage
@@ -83,7 +83,7 @@ def get_configuration():
         for device in data["blockdevices"]:
             if 'children' in device:
                 for child in device['children']:
-                    if child['uuid'] == info["uuid"]:
+                    if child['label'] == info["label"]:
                         #print(f'{os_device} vs {child["path"]}')
                         if os_device == device['path']:
                             cur = child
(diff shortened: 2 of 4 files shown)


packages/ns-storage/files/add-storage
```diff
@@ -28,10 +28,10 @@ mkdir -p "$mdir"
 mount "$device" "$mdir"
 
 # find uuid and create automount config
```
Tbaile, line 30:
```suggestion
# find label and create automount config
```

gsanchietti (approved):

packages/ns-storage/files/ns-storage-check.init
```diff
@@ -0,0 +1,28 @@
...
+# This script is used to disable the storage in case of a disk failure.
+# Must be ran after the fstab init.d script, which is started at 11 and it's name is S11fstab, so this script will be
+# named inside the /etc/rc.d as S11ns-storage-check.
+boot()
+{
+  if [ "$(storage-status)" = "error" ]; then
+    echo "Storage is in error state, disabling write to disk."
+    uci -q delete rsyslog.ns_data
+    uci commit
+  fi
```
gsanchietti, line 19: Suggested changes:
- on error, also remove crontab 
```suggestion
  if [ "$(storage-status)" = "error" ]; then
    echo "Storage is in error state, disabling write to disk."
    uci -q delete rsyslog.ns_data
    uci commit
    crontab -l | grep -v "/usr/sbin/sync-data" | sort | uniq | crontab -
  fi
```

gsanchietti: If the filesystem on /mnt/data is broken, it could be fixed using a a fsck command.
In this case, crontab and syslog must be reconfigured.
So I propose to also add another condition to make sure cron and rsyslog are correctly configured (NOT tested):
```
if [ "$(storage-status)" = "error" ]; then
   if [ -z "$(uci -q get rsyslog.ns_data)" ]; then
        crontab -l | grep -q '/usr/sbin/sync-data' || echo '40 1 * * * /usr/sbin/sync-data' >> /etc/crontabs/root
        uci set rsyslog.ns_data=selector
        uci set rsyslog.ns_data.source='*.*'
        uci set rsyslog.ns_data.destination="/mnt/data/log/messages"
        uci commit rsyslog
   fi
fi
```

The documentation to recovery a broken partition could be something like:
```
fsck.ext4 /dev/sda3
/etc/init.d/ns-storage-check restart
```

gsanchietti (approved):

gsanchietti (approved):

PR #1237, opened 2025-05-27
fix(https-dns-proxy): patched default config, disabling service

Tbaile:
If the package gets installed, on fresh install it overrides the defaults of all DHCP services. This disables the service and adds documentation to enable it correctly.


gsanchietti (approved):
You can go with squash&merge.
Add in the commit message why we moved the package outside the image.

PR #1242, opened 2025-05-29
fix(flashstart): few flashstart adjustments

Tbaile:
Added missing validation for some UI fields
Removed old setup when migrating


gsanchietti (changes requested):
Remove also `flashstart_apply` call from `ns-import` script

gsanchietti (approved):

PR #1275, opened 2025-06-18
fix(ns-api): added support for CIDR in snort exclusions

Tbaile:
Due to a validation issue, the field was not possible to be fillable through the UI, however the tool support the format.


gsanchietti (approved):
Beware of the commit message: it contains `sort` instead of `snort`

PR #1294, opened 2025-07-07
fix(ns-api-server): put unlimited respawn process for ns-api-server

Tbaile:
Discussion: https://mattermost.nethesis.it/nethesis/pl/8byx87muyinw8ms8ermd9py3ue

### DIFF
--- a/packages/ns-api-server/files/ns-api-server.initd
+++ b/packages/ns-api-server/files/ns-api-server.initd
@@ -48,7 +48,7 @@ start_service() {
 
     procd_set_param stdout 1
     procd_set_param stderr 1
-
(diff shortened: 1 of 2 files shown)


gsanchietti (approved):

packages/ns-api-server/files/ns-api-server.initd
```diff
@@ -48,7 +48,7 @@ start_service() {
 
     procd_set_param stdout 1
     procd_set_param stderr 1
-
+    procd_set_param respawn 3600 60 0
```
filippocarletti, line 51: I'd go with default values, I can't see any advantage in retrying forever.

Issue #1244, opened 2025-05-30
qosify configuration persists after WAN removal


filippocarletti:
**Steps to reproduce**
1. Add a WAN connection labeled as "FIBRA100M".
2. Configure qosify for the "FIBRA100M" interface.
3. Remove the "FIBRA100M" WAN from the system.
4. Check the `qosify-status`.

**Expected behavior**
- After removing the "FIBRA100M" WAN, all related qosify configurations should also be removed automatically.

**Actual behavior**
- The qosify configuration remains, even though the "FIBRA100M" interface no longer exists. This is evident from the output of the `qosify-status` command:
  ```
  ===== interface FIBRA100M: not found =====
  ```

**Components**
NethSecurity version: 8-23.05.5-ns.1.4.1

Tbaile:
QA:
Check that configuration related to a deleted RED gets updated correctly.

filippocarletti:
Issue fixed by the new image.
But a new issue arised: Server error in the polict mwan page:
```
root@NethSec:~# /usr/libexec/rpcd/ns.mwan call index_policies
Traceback (most recent call last):
  File "/usr/libexec/rpcd/ns.mwan", line 88, in <module>
    print(json.dumps({'values': mwan.index_policies(e_uci)}))
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/nethsec/mwan/__init__.py", line 343, in index_policies
  File "/usr/lib/python3.11/site-packages/nethsec/mwan/__init__.py", line 306, in __parse_member
uci.UciExceptionNotFound
```

filippocarletti:
Confirm fixed.

PR #1304, opened 2025-07-15
fix(ns-api): removing lingering configuration from QOS and mwan3

Tbaile:
Closes: #1244


packages/ns-api/files/post-commit/clean-network.py
```diff
@@ -0,0 +1,61 @@
...
+
+    for item in changes['network']:
+        # if something in network has been removed
+        if len(item) > 0 and item[0] == 'remove':
+            to_remove_interface = item[1]
+            # if the interface is configured in qosify remove it
+            if to_remove_interface in qosify_interfaces:
+                qosify_changed = True
+                e_uci.delete('qosify', to_remove_interface)
+            # # if the interface is configured in mwan3 remove the interface
```
gsanchietti:
```suggestion
            # if the interface is configured in mwan3 remove the interface
```

packages/ns-api/files/post-commit/clean-network.py
```diff
@@ -0,0 +1,61 @@
...
+                            members_removed.append(entry)
+
+                # iterate over the members removed and remove them from the policies
+                for entry in e_uci.get('mwan3'):
+                    if e_uci.get('mwan3', entry) == 'policy':
+                        members = list(e_uci.get('mwan3', entry, 'use_member', default=[], dtype=str, list=True))
+                        for member in members_removed:
+                            if member in members:
+                                members.remove(member)
+                        e_uci.set('mwan3', entry, 'use_member', members)
```
gsanchietti: To avoid one loop, we could move these lines under the loop on line 40

gsanchietti (approved):

packages/ns-api/files/post-commit/clean-network.py
Tbaile: Not in this case, if we delete a member and we already moved over the policy that contained such member, we skip the de-referencing inside the policy

gsanchietti (commented):
Let's move to pre-commit hooks to avoid a double reload of the config

packages/ns-api/files/pre-commit/clean-network.py
```diff
@@ -0,0 +1,61 @@
...
+                    if e_uci.get('mwan3', entry) == 'policy':
+                        members = list(e_uci.get('mwan3', entry, 'use_member', default=[], dtype=str, list=True))
+                        for member in members_removed:
+                            if member in members:
+                                members.remove(member)
+                        e_uci.set('mwan3', entry, 'use_member', members)
+
+    if qosify_changed:
+        e_uci.save('qosify')
+        e_uci.commit('qosify')
```
gsanchietti: The commit will be done by the API itself
```suggestion
```

packages/ns-api/files/pre-commit/clean-network.py
```diff
@@ -0,0 +1,61 @@
...
+
+    if mwan_changed:
+        e_uci.save('mwan3')
+        e_uci.commit('mwan3')
```
gsanchietti: The commit will be done by the API itself

```suggestion
```

Issue #1310, opened 2025-07-17
feat(ns-plug): add MTU test and configurable MTU field


Tbaile:
**Brief description**

Currently, users experience connectivity issues due to incorrect MTU settings between NethSecurity firewalls and controllers, especially in low-quality network environments. Workarounds like manually setting MTU on the tunnel or using `ifconfig` are not persistent after VPN reconnections, and troubleshooting is difficult without clear visibility into MTU-related problems.

**Proposed solution**

- Integrate an MTU test into the ns-plug client so logs immediately highlight MTU issues, allowing easier diagnosis with a simple grep.
- Add a configurable MTU field to ns-plug's configuration, making MTU adjustments persistent and manageable by users directly from configuration—no manual intervention needed after VPN reconnects.

**Alternative solutions**

- Only expose MTU setting on either the controller or firewall (less flexible, may not resolve all scenarios).
- Use OpenVPN options (`--tun-mtu`, `--fragment`, `--mssfix`) for more dynamic adaptation, but requires additional patching and environment variable support in the VPN container/module.

**Additional context**

- Reference implementation ideas from https://github.com/NethServer/nethsecurity/blob/b30a429b402bacc2615a5fb3726c01235581d2ae/packages/ns-plug/files/ns-plug#L98-L127

**References**
- Helpdesk ticket: https://helpdesk.nethesis.it/a/tickets/191784
- Mattermost Discussion: https://mattermost.nethesis.it/nethesis/pl/qpm7auwxtfn8jeh1i875mmm46o

gsanchietti:
Configured with:
```
uci set ns-plug.config.tun_mtu=1300
uci set ns-plug.config.mssfix=1200
uci commit ns-plug
reload_config
```

Configuration correctly applied:
```
# ip link  | grep tun-nsplug
8: tun-nsplug: <POINTOPOINT,MULTICAST,NOARP,UP,LOWER_UP> mtu 1300 qdisc fq_codel state UNKNOWN mode DEFAULT group default qlen 500

# cat /usr/share/ns-plug/client.conf  | grep -e mss -e mtu
tun-mtu 1300
mssfix 1200
```

Verified

PR #1311, opened 2025-07-17
feat(ns-plug): added custom MTU support

Tbaile:
Also:
- fix(ns-plug): restarting process on postinst and reload

Closes https://github.com/NethServer/nethsecurity/issues/1310


packages/ns-plug/Makefile
```diff
@@ -43,6 +43,7 @@ if [ -z "$${IPKG_INSTROOT}" ]; then
   /usr/libexec/ns-plug/40_ns-plug_automatic_updates
   /etc/init.d/cron restart
   /usr/libexec/ns-plug/40_ns-plug_mwan_hooks
+  /etc/init.d/ns-plug restart
```
gsanchietti: Are you sure this is safe if ns-plug is not running?

packages/ns-plug/files/ns-plug.init
```diff
@@ -32,3 +32,9 @@ service_triggers()
 {
 		procd_add_reload_trigger "ns-plug"
 }
+
+reload_service()
+{
+        stop
+        start
+}
```
gsanchietti, line 40: This should not be needed: who calls the reload?
Also, if the reload is not configured, the fallback is on the restart

Tbaile: This is not true, the reload is being called by:
```
procd_add_reload_trigger "ns-plug"
```

And it's a reload, not a restart, the restart must be forcefully implemented in case of a reload, doc: https://openwrt.org/docs/guide-developer/procd-init-scripts#forcing_service_restart

You can check for this issue by:
- having a connected unit to a controller
- open a terminal and `ubus subscribe service` to listen to events
- change with uci a setting of the ns-plug config, i.e: `uci set ns-plug.config.tls_verify='1' && uci commit ns-plug && reload_config`
- no restart of the service happen, potentially making the connection to the VPN faulty

packages/ns-plug/Makefile
Tbaile: It appears that services get automatically restarted upon install, that didn't happen when we developed `ns-api-server`: https://github.com/NethServer/nethsecurity/blob/9406edb7c1402a692879eefe1e6f0404af8c4430/packages/ns-api-server/Makefile#L54

gsanchietti (approved):

PR #1324, opened 2025-07-29
fix(ns-plug): execute push-info as background job

gsanchietti:
Make sure to not stop OpenVPN connection if push-info fails. Without this fix, the tunnel could be established but not working.


Tbaile (approved):
Really don't like the "throw the process into the abyss" approach, if it really never stops, we'll find ourselves with a bunch of zombie processes. I'd rather shift to a proper "request" through the python package.

But your call!

gsanchietti:
The curl process will die after 60 seconds, while before it could be running for minutes.

PR #1340, opened 2025-08-11
chore(base-files): Better TERM for serial console

filippocarletti:
Use "vt102" for serial console terminal
See https://github.com/openwrt/openwrt/commit/1310e4f1aecef6c2c1b68227216ba7cc04006002 for details.


filippocarletti:
I can't see why vt-102 is better than linux (or xterm).
We could merge this PR just to get the full upstream fix.
Or, better, improve it to run loadkmap only on tty[0-9].

gsanchietti (approved):

PR #1379, opened 2025-09-26
build: restore verbose logs

gsanchietti:
Show the full log directly inside GH actions: no need to download log archive which is difficult to navigate


gsanchietti:
> > which is difficult to navigate
> 
> What did you find hard?

3 main problems:
- I can't see the error from GH action
- the archive does not contain a root dir, so when u unpack it everything it's extract in multiple dirs
- to see the actual error packages has failed, I have to access first the errors.txt then navigate to package log

Issue #1352, opened 2025-08-22
Wireguard Tunnels


Tbaile:
Implement Tunnels as designed by https://github.com/NethServer/nethsecurity/issues/1321.

TODOs:
- [x] Update API to handle automatic updates to the config
- [x] Add Frontend for the feature
- [x] Change `Interfaces` UI to filter out the multiple wireguard  interfaces that will pop up
- [x] Migration of old config
- [x] Periodically refresh wireguard pages

mrmarkuz:
Test case 1 "Fresh install" **verified**

Test case 2 "Update" **verified**

PR #1392, opened 2025-10-07
fix(wireguard): fixing up wireguard implementation

Tbaile:
Fixing issues spotted in QA.
Closes https://github.com/NethServer/nethsecurity/issues/1352


packages/ns-api/files/ns.wireguard
```diff
@@ -200,6 +200,18 @@ def get_peer_defaults(args):
...
+    for server in list_servers():
+        if instance is not None and server['id'] == instance:
+            continue
+        ports_used.append(server['listen_port'])
+    if port in ports_used:
+        return True
+    return False
+
+
+
```
gsanchietti:
```suggestion
```

Issue #1417, opened 2025-11-03
IPS: Configuration fails with CPUs with more than 16 logical cores


Tbaile:
**Steps to reproduce**

- Enable IPS (Snort) for the first time using the Web UI.
- Use a machine with more than 16 logical CPU cores (e.g., 16 physical cores with Hyper-Threading shows 32 logical cores).
- Observe the generated configuration file at /etc/config/snort:
  ```
  config nfq 'nfq'
      option queue_count '32'
      ...
      option thread_count '32'
      ...
  ```
- Try to manually restart Snort (`service snort restart` or `snort-mgr check`).

**Expected behavior**

- Snort should be correctly configured and function with any modern CPU, including those that report more than 16 logical cores.

**Actual behavior**

- Configuration generation sets `queue_count` and `thread_count` to values exceeding the maximum allowed (16), causing configuration errors:
  
  ```
  ERROR: In option queue_count=‘32’, must be 1 <= x <= 16 Errors during generation of snort config
  ```
- The IPS/Snort service silently fails while the Web UI incorrectly indicates a healthy status.

**Components**

- ns-api 3.4.0-r1

**Proposed solution**

- Apply a `min()` function to `queue_count` and `thread_count` values so that they do not exceed 16, regardless of logical core count.

**See also**

- https://community.nethserver.org/t/snort-incorrect-initial-configuration-leading-to-failure/26547

PR #1431, opened 2025-11-14
fix(snort): limit CPU thread and queue count to a maximum of 16

m-dilorenzi:
This pull request introduces a resource management improvement to the Snort setup process. The main change is a safeguard that limits the number of threads and queues used by Snort to a maximum of 16, even if the system has more CPUs available.

Closes: https://github.com/NethServer/nethsecurity/issues/1417


Tbaile (approved):
We agreed not to pursue image update fixes, due to only 2 devices affected by this issue.

Issue #1415, opened 2025-11-03
WireGuard: DNS responses not reaching clients due to incorrect address format


Tbaile:
**Steps to reproduce**

- Set up a WireGuard server tunnel on NethSecurity and assign its DNS server field to the firewall's WireGuard interface address (e.g., 10.189.165.1).
- Connect a WireGuard client and try to resolve DNS using this address.
- Observe that the client does not get responses.

**Expected behavior**

- The WireGuard client should be able to use the assigned DNS server (the firewall’s WireGuard IP) for DNS resolution, as works on OpenWRT.

**Actual behavior**

- DNS responses do not reach the WireGuard client when using the firewall’s WireGuard IP as DNS server.
- Using the LAN IP does not work, and only WAN DNS servers are functional.
- This issue is caused by the automatically generated server IP using an address format that OpenWRT treats as /32, not CIDR (e.g., 10.189.165.1 instead of 10.189.165.1/24).

**Workaround**

Change `addresses 'w.x.y.z/24'` adding /24 where w.x.y.z is the original address on the network configuration

**Components**

- ns-api: 3.4.0-r1

**References**

- https://community.nethserver.org/t/wireguard-wrong-automatically-generated-configuration/26548

PR #1435, opened 2025-11-18
fix(wireguard): patched regression introduced by #1424

Tbaile:
Closes: https://github.com/NethServer/nethsecurity/issues/1415


gsanchietti (commented):
🤷

PR #1453, opened 2025-12-10
build: added softflowd

Tbaile:
Added softflowd package to our repos.

Ref: https://community.nethserver.org/t/some-un-useful-enhancements-for-nethsecurity-tips/26647


gsanchietti (approved):
I think that we should document it, otherwise nobody will know we have it or how to use it

PR #1482, opened 2026-01-20
feat(dpi): added additional logging when dpi blocks

Tbaile:
Following a discussion with @francio87, we identified that support staff face difficulties detecting which applications are being caught by the DPI filter, making troubleshooting more challenging.

Since netifyd v5 now properly supports logging additional output when a block is applied, I've added the log option to the rules. This ensures that logs are saved every 60 seconds in /var/run/netifyd/dpi-actions-*, improving visibility into blocked applications.

### DIFF
--- a/packages/ns-dpi/README.md
+++ b/packages/ns-dpi/README.md
@@ -19,7 +19,7 @@ To enable traffic processing:
 Global options:
 
 - `enabled`: can be `0` or `1`, if set to `1` enable the service
-- `log_blocked`: can be `0` or `1`, if set to `1` blocked connections will be logged
+- `log_blocked`: can be `0` or `1`, if set to `1` blocked connections will be logged inside `/var/log/messages`
 - `firewall_exemption`: can be `0` or `1`, if set to `1` all firewall IP addresses will be
   added to global exemption list and will not match DPI rules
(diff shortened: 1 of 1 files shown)


packages/ns-dpi/README.md
gsanchietti:
```suggestion
- `log_blocked`: can be `0` or `1`, if set to `1` blocked connections will be logged inside `/var/log/messages`
```

packages/ns-dpi/README.md
```diff
@@ -41,6 +41,8 @@ Rule options:
 - `device`: optional device name, if set the rule will be applied only to the given device, example `br-lan`
 - `application`: list of applications to match, the list can contain application names like `netify.amazon-prime`
 - `enabled`: can be `0` or `1`, if set to `1` the rule will be enabled
+- `log`: can be set to `1` to log matching connections, gives an improved visibility on matched connections; 
+  instead of logging a simple message in `/var/log/messages`, logs are stored in `/var/run/netifyd/dpi-actions-*.json`
```
gsanchietti:
```suggestion
  logs are stored in `/var/run/netifyd/dpi-actions-*.json`
```

Issue #1477, opened 2026-01-16
Change default close_action for IPsec connections from "none" to "trap"


cotosso:
### Brief description

Modify the default value of the IPsec parameter `close_action` from `none` to `trap` so that IPsec tunnels are automatically re-established on demand when matching traffic is detected.

### Why

The current default (`close_action=none`) does not automatically re-initiate IPsec tunnels after they go down. This behavior is often unexpected, especially for site-to-site tunnels that administrators assume will come back up automatically when traffic resumes.

As a result, users must manually restart tunnels or add additional logic or monitoring to ensure availability, increasing operational complexity.

### Purpose

Improve the default reliability and usability of IPsec tunnels by ensuring they are automatically reactivated when traffic requires it, without requiring manual intervention or additional configuration.

### Proposed solution

* Change the default value of `close_action` to `trap` for newly created IPsec tunnels.
* Change the default value of `close_action` to `trap` after a tunnel has been edited.

### Alternative solutions

* Expose `close_action` as a selectable option in the UI during tunnel creation, with `trap` preselected.

### Possible negative aspects

* Increased IKE negotiation attempts when matching traffic is generated, potentially increasing background traffic.
* Log noise and repeated connection attempts in case of misconfigured or unreachable peers.
* Unexpected automatic tunnel reactivation for administrators who prefer fully manual control.
* Slight increase in CPU and network usage in setups with many IPsec tunnels and intermittent traffic.

### Components
NethSecurity 8.7.1

Issue #1460, opened 2025-12-29
Support close_action option for IPsec tunnels


cotosso:
**Description**
In NethSecurity 8, the `close_action` parameter of IPsec tunnels is currently set to `none` by default and cannot be managed persistently.

In some scenarios, this behavior prevents the tunnel from being correctly re-established when the remote peer explicitly closes the connection.

It is possible to manually change the value using uci, for example:

```
uci set ipsec.ns_6496322c_tunnel_1.closeaction='start'
``` 

or

```
uci set ipsec.ns_6496322c_tunnel_1.closeaction='trap'
```

However, any subsequent modification of the tunnel configuration from the UI causes this value to be lost and reset to the default.

**Why**
For several site-to-site VPN scenarios, especially when there are multiple SA and the remote peer actively closes the tunnel, the `close_action` parameter is required to ensure automatic re-establishment.
 are
When `close_action` is set to `none`, the tunnel may remain down after a remote-initiated close, reducing availability and requiring manual intervention.

**Proposed solution**
Add support for the `close_action` parameter in the IPsec tunnel configuration so that values set via uci are preserved across configuration changes.

Supported values should include:

* `none`
* `start`
* `trap`

**Future improvement**
Expose the `close_action` parameter in the UI to allow configuration without using uci. This parameter is important in many real-world deployments to ensure tunnel availability.

**Components**
NethSecurity 8.7.1

Issue #1430, opened 2025-11-14
Improve IPsec status visibility for multi-network tunnels


cotosso:
### **Brief description**

When an IPsec tunnel includes multiple networks on the same side (e.g., two or more subnets for the remote site), the UI currently displays only a global connection state (*connected* / *disconnected*).
However, it is possible for one Security Association (SA) to be active while another is not. The UI does not show this granularity.

### **Why / Purpose**

Administrators need to understand the actual operational state of each SA within a multi-network tunnel.
Without visibility on individual SA states, troubleshooting becomes harder, false positives appear (tunnel shown as “connected” even if one SA is down), and administrators must rely on CLI checks instead of the UI.

### **Proposed solution**

- Display per-SA status for each local/remote subnet pair.
Example: a table or expandable panel listing each SA with its state (up / down).

- Provide actions applicable to individual SAs.

**SA-specific actions (examples)**

A well-received enhancement would be to offer SA-level actions directly from the UI. Examples include:

- Restart SA restart only the affected SA without touching others.

- Force rekey : initiate key renegotiation for the selected SA.

- Flush SA : remove the SA from the kernel to force a fresh negotiation.

These actions help identify and fix partial failures more quickly.

Issue #1425, opened 2025-11-13
IPsec: UI edit causes tunnel to not be generated correctly


Tbaile:
**Steps to reproduce**
- Create then modify an IPsec tunnel configuration in NethSecurity
- Observe the tunnel status from the web UI
- Use `swanctl --list-sas --ike <ID>` to examine tunnel session states at the CLI

**Expected behavior**
- When tunnel configuration is modified, changes should be consistently applied to the running system and reflected both in the UI and API states.
- Tunnel should show as "connected" when sessions are established.
- Reloading or reapplying tunnel configuration should not require a full service restart.

**Actual behavior**
- UI shows the tunnel as "not connected" despite the API showing it as "INSTALLED".
- The system uses `grep ESTABLISHED` for session status but the main configuration ID (`ns_d025d5cd` as an example) does not exist, while other configurations use an ID (`ns_51f44731` another example) that does exist. Attempting `swanctl --list-sas --ike <ID>` with the main tunnel's ID fails unless using the correct underlying ID.
- The configuration appears clean, but unless a full service restart (`/etc/init.d/swanctl restart`) is performed, changes are not properly applied and session states are mismatched. This workaround does bring all VPN tunnels down briefly.
- Modifications to tunnels applied with reload do not always fully activate unless a restart is done. Session IDs can persist incorrectly, causing CLI/API/UI mismatches.
- Logs show duplicate CHILD_SA/SPIs and tunnel status confusion, this is probably the cause of the missing re-creation of the new tunnels.

**Workaround**
Restart the swanctl service to refresh strongSwan state:
```
/etc/init.d/swanctl restart
``` 
Note: this will briefly bring down all active VPN tunnels.

**Components**
- ns-api >= 3.4.0

**See also**
- Helpdesk ticket: https://helpdesk.nethesis.it/a/tickets/195376
- Mattermost discussion: https://mattermost.nethesis.it/nethesis/pl/n67dr31ti7ge7p94wrsxi6u8aa

PR #1485, opened 2026-01-21
IPSec: multiple improvements

gsanchietti:
Changes:
- do not delete the tunnel on edit
- change closeaction default value
- allow closeaction custom value
- add extra status info to list-tunnels API
- add restart endpoint

References:
- NethServer/nethsecurity#1425
- NethServer/nethsecurity#1477
- NethServer/nethsecurity#1430
- NethServer/nethsecurity#1460
- https://github.com/NethServer/nethsecurity-ui/pull/676

### DIFF
--- a/packages/ns-api/files/ns.ipsectunnel
+++ b/packages/ns-api/files/ns.ipsectunnel
@@ -126,10 +126,10 @@ def uci_set_if_changed(u, config, section, option, value):
         # Handle list values
         if isinstance(value, list):
             try:
-                current_list = list(u.get_all(config, section, option))
+                current_list = set(sorted(list(u.get_all(config, section, option))))
             except:
                 current_list = None
-            if current_list != value:
+            if current_list != set(sorted(value)):
                 u.set(config, section, option, value)
                 return True
         elif str(current) != str(value):


packages/ns-api/files/ns.ipsectunnel
```diff
@@ -47,14 +47,97 @@ def next_id():
...
+    """Set UCI option only if it differs from current value to minimize changes."""
     try:
-        p = subprocess.run(["swanctl", "--list-sas", "--ike", id], capture_output=True, text=True, check=True)
-        for l in p.stdout.split("\n"):
-            if 'ESTABLISHED' in l:
+        current = u.get(config, section, option, default=None)
+        # Handle list values
+        if isinstance(value, list):
+            try:
+                current_list = list(u.get_all(config, section, option))
```
Tbaile: When updating a UCI property with its current value:
- Non-list values are no-op if unchanged
- List values are removed then re-added regardless

This works, but lists with different ordering are treated as changed. Using a `set` for comparison would solve this and improve performance through hashing.

gsanchietti: Done

Issue #1422, opened 2025-11-06
Snort: home_net configuration does not get updated


francio87:
When the LAN interface network is modified, the `snort.snort.home_net` is not updated.
Even disabling and re-enabling Snort from the UI does not refresh the value.

**Steps to reproduce**

- Modify the LAN interface network settings. (ie change the Lan net from 192.168.1.1/24 to 192.168.50.1/24)
- Check the `snort.snort.home_net` configuration option.
- Disable Snort and re-enable it.

**Expected behavior**

The `snort.snort.home_net` option should update to reflect the new CIDR.

**Actual behavior**

The `snort.snort.home_net` option remains unchanged

**See also**
- https://mattermost.nethesis.it/nethesis/pl/foqxtskjdbnspj33yw6cyiabda

Tbaile:
With [this](https://github.com/NethServer/nethsecurity/pull/1510), the issue is verified.

Issue #1381, opened 2025-09-30
Unify Snort host bypass: one entry for both directions (source and destination)


gsanchietti:
Currently, Snort bypasses are defined separately for source and destination hosts. To bypass traffic to and from a specific host (e.g. google.com), administrators must create two separate entries. This can be confusing and increases the risk of misconfiguration. The feature request proposes unifying bypasses for source and destination into a single host-centric bypass entry.

**Proposed solution**
- Modify the UI and backend to allow creation of a single bypass entry per host, which applies to traffic in both directions (to and from the host).
- Store host bypasses with a single UCI key.
- Update `/usr/share/snort/templates/nftables.uc` to use the unified key and generate both saddr and daddr rules for each bypassed host.
- Rename CLI tooling to include "snort" for clarity (suggestion: `snort-bypass-config` instead of `ns-bypass-config`).
- Migrate existing separate source/destination bypass entries to the new unified format.

**Additional context**
- Reference implementation: `/usr/share/snort/templates/nftables.uc` currently uses keys to generate saddr/daddr rules.
- A single UCI key can be used to generate both rules, simplifying configuration and management.
- CLI tool naming (`ns-bypass-config`) is unclear; including "snort" improves discoverability for administrators.

**See also**
- [Private discussion](https://mattermost.nethesis.it/nethesis/pl/zgpw7ge8efntprrwuttpf6w7qe)

PR #1496, opened 2026-01-29
Snort: new bypass and allow home net configuration

gsanchietti:
Changes:
- Unify logic: having 2 different bypass types can lead to wrong config
- Allow home net configuration

Fixes https://github.com/NethServer/nethsecurity/issues/1422
Implements https://github.com/NethServer/nethsecurity/issues/1381

### DIFF
--- a/packages/ns-api/README.md
+++ b/packages/ns-api/README.md
@@ -8025,16 +8025,17 @@ Response example:
 {
   "enabled": true,
   "ns_policy": "balanced",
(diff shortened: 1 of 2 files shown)


gsanchietti:
Major changes:
- rewritten the migration logic because it didn't take care of bypass with spaces inside description
- improved the api to avoid sending useless data
 
See also the related changes to the UI: https://github.com/NethServer/nethsecurity-ui/pull/683

PR #1497, opened 2026-01-30
perf(netifyd): enabling NFQUEUE

Tbaile:
Boosting Netifyd DPI with nft queue userspace. Refactored to not include DPI changes coming from https://github.com/NethServer/nethsecurity/pull/1520.

Todos:
- if it's possible to dynamically change the nfqueue depending on the cores in the machine, E.g.: If the machine has one core, have just one queue enabled and all traffic should pass there, with two cores distribute load for lan in the first one, then for wan in the second one.
- explore `fanout` option: https://wiki.nftables.org/wiki-nftables/index.php/Queueing_to_userspace

### DIFF
--- a/packages/netifyd/files/etc/netifyd/interfaces.d/10-nfqueue.conf
+++ b/packages/netifyd/files/etc/netifyd/interfaces.d/10-nfqueue.conf
@@ -1,10 +1,4 @@
-# Netify Agent Example Capture Interface Configuration
-# Copyright (C) 2024-2025 eGloo Incorporated
-#
-##############################################################################
-
-# Example Netfilter QUEUE capture source
-##############################################################################
+# NethSecurity default configuration
 
 [capture-interface-LAN]
 capture_type = nfqueue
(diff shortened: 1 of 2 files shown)


packages/netifyd/files/etc/netifyd/interfaces.d/10-nfqueue.conf
```diff
@@ -0,0 +1,23 @@
+# Netify Agent Example Capture Interface Configuration
+# Copyright (C) 2024-2025 eGloo Incorporated
+#
+##############################################################################
+
+# Example Netfilter QUEUE capture source
+##############################################################################
```
gsanchietti: Change to something more useful
```suggestion
#
# NethSecurity default configuration
#
```

packages/netifyd/files/etc/uci-defaults/99-netify-enable-nfqueue.uci-default
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+#
+# Copyright (C) 2025 Nethesis S.r.l.
```
gsanchietti:
```suggestion
# Copyright (C) 2026 Nethesis S.r.l.
```

PR #1505, opened 2026-02-05
chore: revert nginx fork

gsanchietti:
Let's stick to upstream nginx to be sure to gather latest security patches from OpenWrt.
Please note that nginx on OpenWrt platform often lags behind official releases.

After some tests, I've found that the new `ns-reverse-proxy` is enabled on boot on new installations and after an image upgrade.

If we want to update existing machines using packages, we must add ns-reverse-proxy as a dependency of an existing package (eg. ns-plug or ns-api). Merge only when ready to be released

Fix https://github.com/NethServer/nethsecurity/issues/1506

### DIFF
--- a/packages/ns-reverse-proxy/files/ns-reverse-proxy
+++ b/packages/ns-reverse-proxy/files/ns-reverse-proxy
@@ -14,7 +14,6 @@ start_service() {
 		procd_set_param stdout 1
 		procd_set_param stderr 1
 		procd_set_param command '/usr/libexec/nginx-proxy-gen'
-		procd_set_param respawn 3600 60 0
 		procd_close_instance
 }
 
@@ -25,6 +24,5 @@ service_triggers()
 
 reload_service()
 {
-		stop
-        start
+		start
 }


packages/ns-reverse-proxy/files/ns-reverse-proxy
```diff
@@ -0,0 +1,30 @@
...
+START=90
+
+USE_PROCD=1
+
+start_service() {
+		procd_open_instance
+		procd_set_param stdout 1
+		procd_set_param stderr 1
+		procd_set_param command '/usr/libexec/nginx-proxy-gen'
+		procd_set_param respawn 3600 60 0
```
Tbaile: Since it's a trigger process, maybe respawn is not the ideal option

packages/ns-reverse-proxy/files/ns-reverse-proxy
```diff
@@ -0,0 +1,30 @@
...
+}
+
+service_triggers()
+{
+		procd_add_reload_trigger "nginx"
+}
+
+reload_service()
+{
+		stop
```
Tbaile: We can remove the stop, no process will be there anyway

PR #1513, opened 2026-02-10
fix: several packages were disabled after updates

Tbaile:
This will patch out all the flaky startups issues we've faced so far. Even if it needs additional verification and testing for now we put a patch to it to avoid further issues, at another time we'll analyze the problem thoroughly.


Tbaile:
QA: With a fresh image, disable netifyd, flashstart and dpi-license-update services, flash the dev image to check that the services get turned on and enabled.

PR #1520, opened 2026-02-12
refactor(netifyd, ns-dpi): blocking using conntrack labels

### LINKED ISSUE #1526: netifyd: activate nfqueue mode for improved performance
Enable and test netifyd nfqueue mode. The feature is ready but not yet activated after recent package updates. Enabling this mode is expected to improve performance, according to the netifyd producer, potentially by up to 90%.

References:
- https://www.netify.ai/blog/netify-and-netfilter

gsanchietti:
We cannot reliably block applications, seems that different systems behave differently, we are investigating on the why.
Following actions so far have been taken to address this issue:
- removed `nft` set blocks, moving to conntrack labeling approach
- remove netify-init labeling
- use reject instead of drop so the client can stop the requests

Replaces #1519
Closes #1526

### DIFF
--- a/packages/ns-dpi/files/dpi-nft
+++ b/packages/ns-dpi/files/dpi-nft
@@ -21,9 +21,9 @@ chain dpi_actions {
     type filter hook prerouting priority filter + 10; policy accept;
 
     {% if log_enabled -%}
-    ct label netify-block counter log prefix "DPI block: " limit rate {{ log_limit }}
+    ct label netify-blocked counter log prefix "DPI block: " limit rate {{ log_limit }}
     {% endif -%}
-    ct label netify-block counter reject
+    ct label netify-blocked counter reject
     ct label bulk counter ip dscp set cs1 return
     ct label best_effort counter ip dscp set cs0 return
     ct label video counter ip dscp set af41 return
--- a/packages/ns-dpi/files/connlabel.conf
+++ b/packages/ns-dpi/files/connlabel.conf
@@ -1,5 +1,5 @@
 0   netify-init
-1   netify-block
+1   netify-blocked
 2   netify-analyzed
 3   bulk
 4   best_effort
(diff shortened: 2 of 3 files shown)


packages/ns-dpi/files/dpi-nft
```diff
@@ -1,33 +1,51 @@
...
+chain dpi_blocks {
+    type filter hook prerouting priority filter + 10; policy accept;
+
+    # init is to allow kernel to set labels
+    ct label set netify-init
+    {% if log_enabled -%}
+    ct label netify-block counter log prefix "DPI block: " limit rate {{ log_limit }}
+    {% endif -%}
+    ct label netify-block counter drop
+}
```
gsanchietti: Evaluate this changes:
1. if the support does not need to know if a connection has been seen by netifyd, create a dummy chain to initialize the labels
2. use the reject instead of drop so clients will not keep retrying to send TCP packets
```suggestion
chain dpi_dummy {
    # this init is to allow kernel to set labels
    ct label set netify-init
 }
 
chain dpi_blocks {
    type filter hook prerouting priority filter + 10; policy accept;

    {% if log_enabled -%}
    ct label netify-block counter log prefix "DPI block: " limit rate {{ log_limit }}
    {% endif -%}
    ct label netify-block counter reject
}
```

gsanchietti: After few other tests, I'd choose:
- use reject instead of drop
- keep netify-init label so we can easily check which connections have been through netifyd once the nfq is enabled
- re-add removed features to tag a connection with different qos:
  - add back actions inside `dpi-config`
  - add rules inside `dpi-nft`:
    ```
    ct label bulk counter ip dscp set cs1 return
    ct label best_effort counter ip dscp set cs0 return
    ct label video counter ip dscp set af41 return
    ct label voice counter ip dscp set cs6 return
    ```

Given this info, we could improve the conntrack page to show these extra info.
Suggested command:
```
conntrack -L -o labels,id,xml
```

To be used [here](https://github.com/NethServer/python3-nethsec/blob/main/src/nethsec/conntrack/__init__.py#L97C31-L97C59)

PR #1527, opened 2026-02-17
feat(netdata): enable prometheus remote write

gsanchietti:
Tests for Grafana Mimir


gsanchietti:
The snappy library is required but no available inside OpenWrt.
We must switch to something else.

PR #1554, opened 2026-03-04
feat(ns-don): enhance session management with logging and expiration checks

stephdl:
This PR adds time-based session management to the Don remote support tool:

- Tracks session expiration - Sessions now show how much time is left (default 24 hours)
- Auto-cleanup - Hourly cron job automatically stops expired sessions
- Extend command - New don extend adds 7 days (one-time only)
- Logging - All session events (start/stop/extend/expire) now log to syslog
- Bug fixes - Prevents double logging during cleanup and improves variable quoting

Result: Remote support sessions are now time-limited with automatic expiration, better visibility, and audit trails.

```
echo "[4/6] Simulating 25 hours have passed..."
current_time=$(date +%s)
old_time=$((current_time - 90000))  # 90000 seconds = 25 hours
echo $old_time > /var/run/don/session_start
```

```
echo "[7/8] Simulating 8 days have passed..."
current_time=$(date +%s)
old_time=$((current_time - 691200))  # 691200 seconds = 8 days
echo $old_time > /var/run/don/session_start
```

### DIFF
--- a/packages/ns-don/files/don
+++ b/packages/ns-don/files/don
@@ -17,6 +17,10 @@ rsa_key="$base_dir/rsa_key"
 ecdsa_key="$base_dir/ecdsa_key"
 socket="$base_dir/management"
 credentials="$base_dir/credentials"
+session_start="$base_dir/session_start"
+session_extended="$base_dir/session_extended"
+cleanup_lock="$base_dir/.cleanup_lock"
+crontab_file="/etc/crontabs/root"
 user="nethsupport"
 password=""
 ca=$(uci -q get don.config.ca)
@@ -27,8 +31,48 @@ if [ "$2" == "-j" ]; then
     json_output=1
 fi
 
+function log_message
+{
+    # Log to /var/log/messages using logger (syslog)
+    logger -t "don" -p daemon.info "$1"
+}
+
+function log_error
+{
+    # Log errors to syslog
+    logger -t "don" -p daemon.err "$1"
+}
+
+function add_cron_expiration
+{
+    # Add cron job to check expiration every hour
+    if ! grep -q "/usr/sbin/don expire" "$crontab_file" 2>/dev/null; then
+        echo "0 * * * * sleep \$(( RANDOM % 60 )); /usr/sbin/don expire" >> "$crontab_file"
+        [ -x /etc/init.d/cron ] && /etc/init.d/cron restart
+        log_message "Expiration cron job added"
+    fi
+}
+
+function remove_cron_expiration
+{
+    # Remove cron job for expiration checking
+    if [ -f "$crontab_file" ]; then
+        sed -i '\|/usr/sbin/don expire|d' "$crontab_file"
+        [ -x /etc/init.d/cron ] && /etc/init.d/cron restart
+    fi
+}
+
 function cleanup
 {
+    # Use lock file to prevent concurrent cleanup calls
+    if [ -f "$cleanup_lock" ]; then
+        # Another cleanup is already running, skip
+        return
+    fi
+    # Create lock to prevent concurrent executions
+    touch "$cleanup_lock"
+    # Remove expiration cron job
+    remove_cron_expiration
     # stop demons
     [ -f "$sshd_pid" ] && kill "$(cat $sshd_pid)"
     [ -f "$vpn_pid" ] && kill "$(cat $vpn_pid)" && sleep 3 && [ -f "$vpn_pid" ] &&  kill -9 "$(cat $vpn_pid)"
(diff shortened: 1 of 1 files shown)


packages/ns-don/files/don
```diff
@@ -193,7 +242,67 @@ status)
...
+        # Session is still valid
+        hours=$((time_remaining / 3600))
+        minutes=$(( (time_remaining % 3600) / 60 ))
+        
+        if [ "$json_output" == 1 ]; then
+            echo "{\"status\": \"active\", \"expires_in_seconds\": $time_remaining, \"expires_in\": \"${hours}h ${minutes}m\"}"
+        else
+            echo "Session is active - expires in ${hours}h ${minutes}m"
+        fi
+        exit 0
```
gsanchietti: No need for this else branch, we already have the same logic using "don status"

packages/ns-don/files/don
```diff
@@ -193,7 +242,67 @@ status)
 	exit 1
     fi
 ;;
+check-expiry)
+    # Check if session has expired
+    if [ ! -f "$session_start" ]; then
```
gsanchietti: Do not use $session_start as flag to know if a session is started: if a session is present and the package is updated, the session_start files is not created.
So for check expire combine: both $session_start and $credentials must be present

packages/ns-don/files/don
```diff
@@ -193,7 +242,67 @@ status)
...
+    old_remaining=$((old_expiry - current_time))
+    
+    # Reset start time to NOW
+    date +%s > $session_start
+    
+    # Log the extension (without sensitive data)
+    log_message "Remote support session extended by 24 hours"
+    
+    echo "Session extended by 24 hours"
+    /usr/sbin/don status
```
gsanchietti: Let's mimic the NethServer behavior: extend the sessions to 7 days. After 7 days the session is always stopped.

packages/ns-don/files/don
```diff
@@ -193,7 +242,67 @@ status)
...
+        
+        if [ "$json_output" == 1 ]; then
+            echo "{\"status\": \"active\", \"expires_in_seconds\": $time_remaining, \"expires_in\": \"${hours}h ${minutes}m\"}"
+        else
+            echo "Session is active - expires in ${hours}h ${minutes}m"
+        fi
+        exit 0
+    fi
+;;
+extend)
```
Tbaile, line 332: Could be possible to add a way to know if the session has been extended? It will be beneficial to the UI that will show another message

packages/ns-don/files/don
```diff
@@ -193,7 +242,67 @@ status)
...
+        exit 1
+    fi
+    
+    old_start=$(cat $session_start)
+    old_expiry=$((old_start + 86400))
+    current_time=$(date +%s)
+    old_remaining=$((old_expiry - current_time))
+    
+    # Reset start time to NOW
+    date +%s > $session_start
```
Tbaile: We use so many variables and we just restart the timer? Are these leftovers from some tests or the restart is the wanted approach?

packages/ns-don/files/don
stephdl: done thanks

packages/ns-don/files/don
stephdl: I add a file with the timestamp of the end of session

packages/ns-don/files/don
stephdl: fixed

packages/ns-don/files/don
stephdl: we only extend to (now + (7*24h))

packages/ns-don/files/don
Tbaile, line 1: With these changes the UI breaks, badly
the returns for the start and status calls are doubled json objects, which break the backend and the UI is not able to show the state
```
root@NSEC-Tommaso:~# don start -j
{"server_id": "XXXX", "session_id": "XXXX"}
{"expires_in_seconds": 86400, "expires_in": "24h 0m"}
```

The status call returns null if session is not in place, having then issues figuring out what is the state
```
root@NSEC-Tommaso:~# don status
don is not running
root@NSEC-Tommaso:~# don status -j
null
```

I suggest:
- single json per command, keeping the old: `{"server_id": "XXX", "session_id": "XXX"}` and adding field to this one, avoiding breaking changes
- for don status and start return (if called using the `-j` option) the expiration timestamp, then the UI can manage to show the correct time from there

packages/ns-don/files/don
```diff
@@ -193,7 +254,58 @@ status)
 	exit 1
     fi
 ;;
+check-expiry)
```
gsanchietti: The action name `check-expiry` seems that only checks and not actually does any work.
I suggest to change it to `expire`

packages/ns-don/files/don
```diff
@@ -193,7 +254,58 @@ status)
...
+    
+    # Set expiry to 7 days from now
+    current_time=$(date +%s)
+    echo $((current_time + 604800)) > $session_extended
+    
+    # Log the extension (without sensitive data)
+    log_message "Remote support session extended by 7 days"
+    
+    echo "Session extended by 7 days"
+    /usr/sbin/don status
```
gsanchietti: I see 2 problems here:
- we do not need to print the status output here
- we need to output json if called by the UI

I suggest something like this:
```suggestion
    
    if [ "$json_output" == 1 ]; then
            echo "{\"status\": \"extended\", \"expiration\": \"....\"}"
        else
            echo "Session extended by 7 days"
    fi
```

packages/ns-don/files/don
```diff
@@ -27,8 +30,49 @@ if [ "$2" == "-j" ]; then
...
+    [ -n "$system_id" ]
+}
+
+function add_cron_expiration
+{
+    # Add cron job to check expiration every hour (only if subscription exists)
+    if has_subscription; then
+        if ! grep -q "/usr/sbin/don expire" "$crontab_file" 2>/dev/null; then
+            echo "0 * * * * sleep \$(( RANDOM % 60 )); /usr/sbin/don expire" >> "$crontab_file"
+            log_message "Expiration cron job added"
```
gsanchietti: This looks like a debug, let's remove it:
```suggestion
```

packages/ns-don/files/don
```diff
@@ -27,8 +30,49 @@ if [ "$2" == "-j" ]; then
...
+        fi
+    fi
+}
+
+function remove_cron_expiration
+{
+    # Remove cron job for expiration checking
+    if [ -f "$crontab_file" ]; then
+        sed -i '\|/usr/sbin/don expire|d' "$crontab_file"
+        log_message "Expiration cron job removed"
```
gsanchietti: This looks like a debug, let's remove it:
```suggestion
```

packages/ns-don/files/don
```diff
@@ -40,19 +84,57 @@ function cleanup
...
+    local current_time
+    current_time=$(date +%s)
+    local time_remaining=$((session_expiry - current_time))
+    local hours_remaining=$((time_remaining / 3600))
+    local minutes_remaining=$(( (time_remaining % 3600) / 60 ))
+
     if [ "$json_output" == 1 ]; then
-        echo "{\"server_id\": \"$(sed -n '1{p;q}' $credentials)\", \"session_id\": \"$(sed -n '2{p;q}' $credentials)\"}"
+        if [ "$time_remaining" -gt 0 ]; then
+            echo "{\"server_id\": \"$server_id\", \"session_id\": \"$session_id\", \"expires_in_seconds\": $time_remaining, \"expires_in\": \"${hours_remaining}h ${minutes_remaining}m\"}"
```
gsanchietti: For the UI, `exipres_in_seconds` is enough so I'd remove `expires_in` from the JSON

packages/ns-don/files/don
```diff
@@ -40,19 +84,57 @@ function cleanup
...
+        else
+            echo "{\"server_id\": \"$server_id\", \"session_id\": \"$session_id\", \"status\": \"expired\"}"
+        fi
     else
-        echo -n "Server ID:"
-        echo -e "\t"$(sed -n '1{p;q}' $credentials)
-        echo -n "Session ID:"
-        echo -e "\t"$(sed -n '2{p;q}' $credentials)
+        echo "Server ID:$(printf '\t')$server_id"
```
gsanchietti, line 138: You can move the calculation of hours and minutes here, it can be done only when we need to display the info on the terminal.

packages/ns-don/files/don
```diff
@@ -162,38 +246,108 @@ AuthorizedKeysFile $auth_keys
 StrictModes no
 EOF
 
-    mkdir -m 0700 -p /var/empty
-    /usr/sbin/sshd -f $sshd_conf
-    cat $(uci -q get don.config.ssh_key) > $auth_keys
+    mkdir -p /var/empty
+    chmod 700 /var/empty
+    /usr/sbin/sshd -f "$sshd_conf"
+    cat "$(uci -q get don.config.ssh_key)" > "$auth_keys"
```
gsanchietti, line 259: Why this has been changed?

packages/ns-don/files/don
```diff
@@ -162,38 +246,108 @@ AuthorizedKeysFile $auth_keys
...
+        fi
+        log_error "Remote support session expired - auto-disabling"
+        /usr/sbin/don stop
+        exit 0
+    fi
+;;
+extend)
+    # Extend session by 7 days
+    if [ ! -f "$session_start" ] || [ ! -f "$credentials" ]; then
+        echo "No session running"
```
gsanchietti: This will break the UI

packages/ns-don/files/don
stephdl: shellcheck warning IIRC

packages/ns-don/files/don
```diff
@@ -162,38 +236,112 @@ AuthorizedKeysFile $auth_keys
...
+    if [ $time_remaining -le 0 ]; then
+        # Session has expired - STOP IT NOW
+        if [ "$json_output" == 1 ]; then
+            echo "{\"status\": \"expired\", \"action\": \"stopping\"}"
+        else
+            echo "Session expired - stopping don service"
+        fi
+        log_error "Remote support session expired - auto-disabling"
+        /usr/sbin/don stop
+        exit 0
```
gsanchietti: The current command does not cleanup the session.
```suggestion
        cleanup
```

packages/ns-don/files/don
```diff
@@ -162,38 +254,118 @@ AuthorizedKeysFile $auth_keys
...
+            echo "Session already extended"
+            echo ""
+        fi
+        get_session_info "already_extended"
+        exit 0
+    fi
+
+    # Extend: store 24h + 7 days as total duration relative to session_start
+    echo "$((86400 + 604800))" > "$session_extended"
+
```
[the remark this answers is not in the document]
stephdl: @copilot apply changes based on [this feedback](https://github.com/NethServer/nethsecurity/pull/1554#discussion_r3008571548)

PR #1556, opened 2026-03-05
feat: restore extra packages

gsanchietti:
After an image upgrade, try to restore extra installed packages

Implementation details:
- add a new binary and uci default only inside the image
- start a task in background that tries to reinstall previously installed packages: the script logs what it does and terminates after 5 minutes if the WAN was not ready


files/etc/init.d/ns-restore-extra-packages
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh /etc/rc.common
+#
+# Copyright (C) 2026 Nethesis S.r.l.
+# SPDX-License-Identifier: GPL-2.0-only
+#
+# Procd init script to restore extra packages after a system upgrade.
+# Runs late at boot (START=99), waits for WAN inside the script itself.
+
+START=99
```
gsanchietti, line 10: Remember to add a restart here

PR #1566, opened 2026-03-12
feat(ovpnrw): connections history from storage if available

### LINKED ISSUE #1404: Show full OpenVPN RW in the UI connection history from stored logs instead of RAM
**Description:**
Currently, the UI displays the OpenVPN Roadwarrior connection history only since the last firewall reboot. This happens because the data is read from memory (RAM), which is cleared at each restart.

**Why / Purpose**
Administrators need visibility on the complete connection history to audit remote access and troubleshoot VPN usage over time, regardless of system uptime. Limiting the view to the current session hinders analysis and reporting.

**Proposed solution**
Read the OpenVPN connection history from the persistent logs stored in the system storage (`/mnt/data/openvpn/ns_roadwarrior1/connections.db`) instead of relying only on data kept in RAM. This would allow the UI to show all past connections, not just those after the last reboot.

m-dilorenzi:
This pull request introduces several improvements and enhancements to the OpenVPN connections history section, primarily focused on supporting persistent storage for connection history databases, improving API functionality, and adding mechanisms to synchronize connection records when storage is mounted.
The changes ensure that connection history is preserved across reboots, provide more flexible API filtering and sorting, and add scripts to automate database migration.

Persistent storage and database path handling:

* All relevant code now checks for the presence of `/mnt/data/openvpn` and uses it as the preferred location for `connections.db`, falling back to `/var/openvpn` if storage is not available. Helper functions like `get_connections_db_path` and `get_ovpnrw_db_path` were introduced to centralize this logic.
* The documentation (`README.md`) was updated to clarify that connection history is stored in either `/var/openvpn/<instance>/connections.db` or `/mnt/data/openvpn/<instance>/connections.db` depending on storage configuration.

Database migration and archiving:

* A new hotplug script (`20-openvpn-merge-connections-db`) and migration utility (`openvpn-merge-connections-db`) were added to synchronize connection records from `/var` to `/mnt/data` when storage is mounted, ensuring persistence across reboots.
* When removing an OpenVPN instance, the connection history database is now archived before deletion to prevent data loss.

API enhancements and reporting improvements:

* The `connection_history` API was refactored to support filtering (by time range, accounts, and search query), sorting (including manual sorting for IP addresses), and pagination, returning richer metadata for frontend use.
* All reporting functions now use the new database path resolution logic, ensuring data is pulled from the correct location regardless of storage configuration.

These changes collectively improve reliability, flexibility, and maintainability of the OpenVPN connections history section.

Closes: https://github.com/NethServer/nethsecurity/issues/1404

### DIFF
--- a/packages/ns-api/files/ns.ovpnrw
+++ b/packages/ns-api/files/ns.ovpnrw
@@ -17,11 +17,12 @@ import ipaddress
 import subprocess
 from euci import EUci
 from nethsec import utils, ovpn, firewall, users, objects
+import datetime as _datetime
 from datetime import datetime, timezone
 import tarfile
 import io
 import sqlite3
-import time
+import time as _time
 import datetime
 import csv
 from zoneinfo import ZoneInfo
@@ -230,6 +231,32 @@ def get_user_extra_config(u, user_id):
     except:
         return {}
 
+def archive_connections_db(instance):
+    timestamp = _datetime.datetime.now().strftime('%Y%m%d%H%M%S')
+    var_db = f'/var/openvpn/{instance}/connections.db'
+    storage_db = f'/mnt/data/openvpn/{instance}/connections.db'
+    # archive /var DB
+    if os.path.exists(var_db):
+        archive_path = var_db.replace('connections.db', f'connections-{timestamp}.db')
+        try:
+            shutil.move(var_db, archive_path)
+        except:
+            pass
+    # archive /mnt/data DB if it exists
+    if os.path.exists(storage_db):
+        archive_path = storage_db.replace('connections.db', f'connections-{timestamp}.db')
+        try:
+            shutil.move(storage_db, archive_path)
+        except:
+            pass
+
+def get_connections_db_path(ovpninstance):
+    storage_db = f'/mnt/data/openvpn/{ovpninstance}/connections.db'
+    var_db = f'/var/openvpn/{ovpninstance}/connections.db'
+    if os.path.isdir('/mnt/data') and os.path.exists(storage_db):
+        return storage_db
+    return var_db
+
 ## APIs
 
 def list_bridges():
@@ -356,6 +383,9 @@ def remove_instance(instance):
         remove_tap_from_bridge(u, old_bridge, old_dev)
 
     u.delete("openvpn", instance)
+    
+    # archive connection history databases before removing the instance directory
+    archive_connections_db(instance)
     shutil.rmtree(f"/etc/openvpn/{instance}", ignore_errors=True)
     try:
         # workaround: manually delete config since it's not removed from init script
@@ -665,7 +695,6 @@ def disconnect_user(ovpninstance, username):
         return utils.generic_error("user_disconnect_failed")
     return {"result": "success"}
 
-
 def disable_user(ovpninstance, username):
     u = EUci()
     try:
@@ -966,7 +995,7 @@ def download_user_2fa(ovpninstance, username):
         return utils.validation_error("username", "2fa_download_failed", username)
 
 def connection_history_csv(ovpninstance, timezone):
-    database_path = f'/var/openvpn/{ovpninstance}/connections.db'
+    database_path = get_connections_db_path(ovpninstance)
 
     try:
         conn = sqlite3.connect(database_path)
@@ -1040,9 +1069,32 @@ def connection_history_csv(ovpninstance, timezone):
     # Return the path of the CSV file
     return {"csv_path": csv_file_path}
 
-
-def connection_history(ovpninstance):
-    database_path = f'/var/openvpn/{ovpninstance}/connections.db'
+def connection_history(ovpninstance, q='', time_range='', accounts=None, page=1, per_page=10, sort_by='startTime', desc=True):
+    database_path = get_connections_db_path(ovpninstance)
+    
+    # map frontend field names to database column names
+    sort_field_mapping = {
+        'account': 'common_name',
+        'startTime': 'start_time',
+        'endTime': 'start_time + duration',
+        'duration': 'duration',
+        'virtualIpAddress': 'virtual_ip_addr',
+        'remoteIpAddress': 'remote_ip_addr'
+    }
+    
+    # default to startTime if sort_by is invalid
+    if sort_by not in sort_field_mapping:
+        sort_by = 'startTime'
+    
+    # determine if we need to sort in Python (for IPs) or SQL (for other fields)
+    sort_manually = sort_by in ['virtualIpAddress', 'remoteIpAddress']
+    
+    if not sort_manually:
+        sort_column = sort_field_mapping[sort_by]
+        sort_order = 'DESC' if desc else 'ASC'
+    else:
+        sort_column = None
+        sort_order = None
 
     try:
         conn = sqlite3.connect(database_path)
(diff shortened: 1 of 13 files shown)


packages/ns-api/files/ns.ovpnrw
```diff
@@ -17,11 +17,12 @@ import ipaddress
 import subprocess
 from euci import EUci
 from nethsec import utils, ovpn, firewall, users, objects
+import datetime as _datetime
 from datetime import datetime, timezone
 import tarfile
 import io
 import sqlite3
-import time
+import time as _time
```
Tbaile: Something is wrong here, we have multiple and renamed imports.
Some come from before, adjust them if possible.

packages/ns-openvpn/files/openvpn-merge-connections-db
```diff
@@ -0,0 +1,113 @@
...
+#
+# Copyright (C) 2026 Nethesis S.r.l.
+# SPDX-License-Identifier: GPL-2.0-only
+#
+
+# Merge OpenVPN connection records from /var into storage (/mnt/data).
+# Called on storage mount (via add-storage script on /usr/sbin/add-storage)
+
+import os
+import shutil
```
Tbaile: Hard pass on using this library for copy and copy2, from the docs:
> Warning Even the higher-level file copying functions ([shutil.copy()](https://docs.python.org/3/library/shutil.html#shutil.copy), [shutil.copy2()](https://docs.python.org/3/library/shutil.html#shutil.copy2)) cannot copy all file metadata.
On POSIX platforms, this means that file owner and group are lost as well as ACLs.

packages/ns-openvpn/files/80-save-connection
```diff
@@ -12,6 +12,14 @@ import sqlite3
 from euci import EUci
 from nethsec import users
 
+STORAGE_BASE = '/mnt/data'
+VAR_BASE = '/var'
+
+def get_db_path(instance):
+    if os.path.isdir(STORAGE_BASE):
```
Tbaile: These checks are risky, better stick to the source of truth:
```
uci get fstab.ns_data.target='/mnt/data'
```

packages/ns-openvpn/files/20-openvpn-merge-connections-db
```diff
@@ -0,0 +1,22 @@
...
+#
+# Copyright (C) 2026 Nethesis S.r.l.
+# SPDX-License-Identifier: GPL-2.0-only
+#
+# Triggered by hotplug when a block device is added.
+# If /mnt/data is mounted, sync OpenVPN connection records from /var to storage.
+
+[ "$ACTION" = "add" ] || exit 0
+[ "$DEVTYPE" = "partition" ] || exit 0
+[ -x /usr/libexec/ns-openvpn/openvpn-merge-connections-db ] || exit 0
```
Tbaile: This is not needed

packages/ns-openvpn/files/20-openvpn-merge-connections-db
```diff
@@ -0,0 +1,22 @@
...
+
+# Wait for /mnt/data to be mounted (up to 10 seconds)
+i=0
+while [ $i -lt 10 ] && ! grep -q ' /mnt/data ' /proc/mounts; do
+    sleep 1
+    i=$((i + 1))
+done
```
Tbaile, line 26: If it's not mounted, what do we do? seems we just exit with 0 anyway
It the drive gets mounted but it's not configured we just fail, right?

packages/ns-api/files/ns.ovpnrw
```diff
@@ -1053,10 +1105,78 @@ def connection_history(ovpninstance):
...
+            params.append(int(_datetime.datetime.combine(first_of_last_month, _datetime.time.min).timestamp()))
+        elif time_range == 'last_3_months':
+            today = _datetime.date.today()
+            month = today.month - 3
+            year = today.year
+            if month <= 0:
+                month += 12
+                year -= 1
+            first_of_3_months_ago = _datetime.date(year, month, 1)
+            conditions.append('start_time >= ?')
```
Tbaile, line 1101: This is a bit error prone, a more elegant solution should be that the UI passes to the backend a start and end timedate in UTC format, then the backend simply filters the data by what the frontend passed.

This is a mere suggestion

Tbaile (changes requested):

packages/ns-openvpn/files/20-openvpn-merge-connections-db
m-dilorenzi: Why should we remove these checks? They ensure the merge script runs only when a new partition is added, preventing it from executing during potentially dangerous operations like removing or updating storage.

Tbaile: Not the first two, only the one regarding `openvpn-merge-connections-db`, the file will always be there, and always be executable as you wrote down in the makefile

packages/ns-openvpn/files/20-openvpn-merge-connections-db
m-dilorenzi: If after 10 seconds the storage has not been mounted and configured, merge will not be executed and a warning message is logged.

packages/ns-api/files/ns.ovpnrw
m-dilorenzi: The UI will now send a new parameter called `start_time` with the unix timestamp to use for the database query, avoiding calculate the timestamp here in the backend.

packages/ns-openvpn/files/80-save-connection
```diff
@@ -12,6 +12,24 @@ import sqlite3
...
+    except:
+        pass
+    return None
+
+def get_db_path(instance):
+    storage = get_storage_base()
+    if storage:
+        return os.path.join(storage, 'openvpn', instance, 'connections.db')
+    return os.path.join(VAR_BASE, 'openvpn', instance, 'connections.db')
+
```
Tbaile:
```suggestion
def get_db_path(instance):
    u = EUci()
    base_path = u.get('fstab', 'ns_data', 'target', default='')
    if not os.path.isdir(base_path):
        base_path = '/var'
    return os.path.join(base_path, 'openvpn', instance, 'connections.db')

```

Isn't something like this simpler? Is there a need for what you did there?

packages/ns-openvpn/files/80-save-disconnection
```diff
@@ -23,6 +42,18 @@ bytes_sent = int(env.get('bytes_sent', '0'))
...
+    # insert a new full record with all the data on /var
+    start_time = int(env.get('time_unix', '0'))
+    virtual_ip_addr = env.get('ifconfig_pool_remote_ip', '')
+    remote_ip_addr = env.get('untrusted_ip', '')
+    c.execute(
+        "INSERT INTO connections (common_name, virtual_ip_addr, remote_ip_addr, start_time, duration, bytes_received, bytes_sent) "
+        "VALUES (?, ?, ?, ?, ?, ?, ?)",
+        (common_name, virtual_ip_addr, remote_ip_addr, start_time, duration, bytes_received, bytes_sent)
+    )
+
```
Tbaile: Maybe this can be simplified with an upsert: https://sqlite.org/lang_upsert.html

packages/ns-openvpn/files/openvpn-merge-connections-db
Tbaile, line 1: This too, maybe an UPSERT is the better choice without the need for checks on present data

packages/ns-openvpn/files/init-connections-db
```diff
@@ -1,12 +1,23 @@
...
+
+def get_storage_base():
+    try:
+        u = EUci()
+        target = u.get('fstab', 'ns_data', 'target')
+        if target and os.path.isdir(target):
+            return target
+    except:
+        pass
+    return None
```
Tbaile: This function can be easily be removed and the logic being put in the call below, simplification can be done by doing what I've suggested on the ns.report

Tbaile (changes requested):
There's plenty to optimize yet, but after this I think we're done, this seems to be working well.

[PyUCI](https://pypi.org/project/pyuci/) gives plenty of helpers to retrieve data, so we don't actually have to constantly check for null values.

packages/ns-openvpn/files/openvpn-merge-connections-db
m-dilorenzi: Using the UPSERT query requires a unique constraint on at least one database field. Since we don't have any unique constraints, we cannot use UPSERT.

packages/ns-openvpn/files/80-save-disconnection
m-dilorenzi: Using the UPSERT query requires a unique constraint on at least one database field. Since we don't have any unique constraints, we cannot use UPSERT.

packages/ns-openvpn/files/init-connections-db
Tbaile: This is the only thing that needs a patch like the one you've made above, everything else checks out!

Tbaile (changes requested):

PR #1578, opened 2026-03-31
fix(flashstart): proplus fixes

Tbaile:
Flashstart now skips ips on bypass instead of routing them all inside the redirect chain.
Solved issue where ranges of IPs were not applied correctly into the firewall.


gsanchietti (approved):
🤷‍♂️

PR #1581, opened 2026-04-01
fix(ns-don): updated expiry methods, some minor adjustments

Tbaile:
Updated some of the behaviours of the script, especially on expiration:
- moving the cron every 5 minutes, instead of start of hour
- returning the UTC timestamp of the expiration, not anymore using the start of the session

### DIFF
--- a/packages/ns-don/files/don
+++ b/packages/ns-don/files/don
@@ -45,9 +45,9 @@ function log_error
 
 function add_cron_expiration
 {
-    # Add cron job to check expiration every minute
+    # Add cron job to check expiration every 10 minutes
     if ! grep -q "/usr/sbin/don expire" "$crontab_file" 2>/dev/null; then
-        echo "*/5 * * * * /usr/sbin/don expire" >> "$crontab_file"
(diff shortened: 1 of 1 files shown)


packages/ns-don/files/don
```diff
@@ -45,9 +45,9 @@ function log_error
 
 function add_cron_expiration
 {
-    # Add cron job to check expiration every hour
+    # Add cron job to check expiration every minute
```
gsanchietti:
```suggestion
    # Add cron job to check expiration every 10 minutes
```

packages/ns-don/files/don
```diff
@@ -45,9 +45,9 @@ function log_error
...
     if ! grep -q "/usr/sbin/don expire" "$crontab_file" 2>/dev/null; then
-        echo "0 * * * * sleep \$(( RANDOM % 60 )); /usr/sbin/don expire" >> "$crontab_file"
+        echo "*/5 * * * * /usr/sbin/don expire" >> "$crontab_file"
```
gsanchietti: Every 5 minutes is too much
```suggestion
        echo "*/10 * * * * /usr/sbin/don expire" >> "$crontab_file"
```

PR #1592, opened 2026-04-13
fix(flashstart): avoiding bypass redirect, removing from ipsets

### LINKED ISSUE #1579: flashstart: ipsets ranges not added and bypasses not working correctly
**Steps to reproduce**

Deploy the current NethSecurity system with FlashStart enabled.

**Expected behavior**

IP sets and bypass mechanisms should be added and function as expected without errors.

**Actual behavior**

If IP sets contain a range of IPs (using '-') they are not correctly added resulting in not sending dns requests where they should. Additionally if the IP is in the bypass and on a profile, it will get redirected anyway to flashstart, causing issues.

**Components**

ns-flashsstart `1.0.3-r2`

Tbaile:
Seems that the fix introduced in https://github.com/NethServer/nethsecurity/pull/1578 had the downside to always redirects the requests. This gives the ick to mail servers that expect that the queries is not getting tampered in **any possible way**.

This fix removes the bypasses entirely from the records being sent from flashstart, allowing to escape firewall redirect entirely and to fall down in rules and everything else.

Closes https://github.com/NethServer/nethsecurity/issues/1579

### DIFF
--- a/packages/ns-flashstart/files/ns-flashstart
+++ b/packages/ns-flashstart/files/ns-flashstart
@@ -48,6 +48,36 @@ def __expand_ip_range(entry: str) -> list:
     return [entry]
 
 
+def __load_bypass_rules() -> list:
+    rules = []
+    for entry in e_uci.get('flashstart', 'global', 'bypass', list=True, dtype=str, default=[]):
+        if entry == '':
+            continue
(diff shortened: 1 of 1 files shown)


gsanchietti (commented):
🤷

PR #1609, opened 2026-04-23
feat(migration): move to new my

edospadoni:
## Summary

Final migration PR that takes the appliance off `backupd.nethesis.it`, off my-old (`/api/`, `/isa/`) and off the `my.nethesis.it/proxy/*` translation layer, and moves it onto the `my` collect API with its own native credentials. After this ships, `type=enterprise` units talk directly to the my collect endpoints; `type=community` units are left on the legacy `my.nethserver.com` / `backupd.nethesis.it` infrastructure — that is explicitly out of scope for the my migration.

### What it does

- **`migrate-to-my`** (new idempotent one-shot, enterprise-only). Runs at the top of every send-* script.
  1. Calls the translation proxy's `/proxy/credentials` with the legacy Basic-Auth pair.
  2. Atomically rotates `ns-plug.config.system_id` / `secret` to the mapped my credentials.
  3. Preserves the legacy pair under `legacy_system_id` / `legacy_secret` (audit + rollback).
  4. (Re)asserts `collect_url`, stamps `migrated_at` and sets `migrated='1'` so subsequent runs are no-ops.

- **Scripts cutover**. `send-heartbeat`, `send-inventory`, `send-backup`, `remote-backup` all drop the dual-send to backupd + proxy and do a single call to `$collect_url/{heartbeat,inventory,backups/*}` with rotated credentials on the enterprise branch. Community branches are byte-for-byte untouched.

- **Native registration**. `/usr/sbin/register` enterprise branch now POSTs `my.nethesis.it/backend/api/systems/register` with `{system_secret: <pasted>}`, reads back `system_key`, writes `collect_url` + `migrated='1'` so a fresh unit lands on the new my without passing through the proxy. Community register is unchanged.

- **Subscription info read path**. `/usr/sbin/subscription-info` enterprise branch now queries collect `/info` with the rotated credentials and emits a legacy-shaped envelope so `ns.subscription info` and the existing Vue UI keep parsing the same keys. Plan / expiration fields fall back to the organization name / null — the new my data model no longer tracks a subscription plan at the system level.

- **Migration fingerprint in phonehome**. `phonehome` now publishes a `facts.migration` block with:
  - `from_legacy_system_id`: the pre-rotation UUID when the unit was migrated through the proxy, `null` on fresh natively-registered units.
  - `migrated_at`: ISO 8601 timestamp of the rotation, `null` on fresh units.
  
  This lets `my` count how many legacy units have completed the rotation and decide when `backupd.nethesis.it` and the translation proxy can be decommissioned.

### Resilience

A `/proxy/credentials` outage during an enterprise unit's upgrade window leaves the unit on legacy credentials against collect, which returns `401`. `migrate-to-my` is re-invoked every 10 minutes via `send-heartbeat`'s cron entry, so the unit recovers automatically once the proxy is back up. Accepted trade-off: no dual-mode in the scripts; the simpler single-send path is preferred.

### Blocking / coordination

> [!WARNING]
> **Do not merge this PR until `my` is in production.**
> Once this ships, enterprise appliances stop talking to `backupd.nethesis.it` and my-old entirely. If `my` is not serving the `/collect/api/systems/*` endpoints and the `/proxy/credentials` rotation helper in production, the unit goes silent from `my`'s perspective until the services come up.

Requires in order:
1. NethServer/my#81 (my backend + collect + UI + security hardening + DELETE route) merged and deployed to production.
2. #1608 (dual-send) merged first, so this PR's diff stays minimal.
3. This PR merged and released.
4. NethServer/nethsecurity-ui#746 (Vue UI consumption of the my-native JSON) merged immediately after.

### Post-merge operational follow-up

- Monitor `/proxy/credentials` success rate during the upgrade rollout.
- On the my admin side, count inventories where `facts.migration.from_legacy_system_id != null`. When this count equals the rows in `proxy_mappings`, every legacy unit has rotated and `backupd.nethesis.it` + the translation proxy can be decommissioned.

Tracking issue: NethServer/my#83.


edospadoni:
Added the **hotspot login via My Nethesis** (`d7d2d01c`): `ns.dedalo` gains `oidc-start`/`oidc-poll` implementing the device pairing against the hotspot manager (nethesis/icaro#200); remote curl calls now have timeouts; the linked account name/user are stored in uci and cleared on password login/unregister. The ns-ui pin is bumped to the UI counterpart (`607826e3` → NethServer/nethsecurity-ui#746 @9209ee9). NOTE: the OIDC login works only once icaro PR #200 is deployed on my.nethspot.com — until then the UI reports it as not supported and the legacy password login keeps working.

packages/victoria-metrics/files/vmalert.initd
```diff
@@ -19,25 +19,37 @@ start_service() {
...
     }
 
+    notifier_url=""
     if [ "$system_type" = "enterprise" ] && [ -n "$system_id" ] && [ -n "$system_secret" ]; then
-        notifier_url="https://my.nethesis.it/proxy/alerts"
+        if [ "$migrated" = "1" ] && [ -n "$collect_url" ]; then
+            # Native my Mimir alertmanager, derived from collect_url, e.g.
+            #   https://my-proxy-prod.onrender.com/collect/api/systems
+            #   -> https://my-proxy-prod.onrender.com/collect/api/services/mimir/alertmanager
+            notifier_url="${collect_url%%/collect/*}/collect/api/services/mimir/alertmanager"
```
Tbaile: Variable sub is always nasty on initd, try to keep away from it, a second uci variable is more suited

packages/ns-phonehome/files/phonehome
Tbaile, line 1: We can see the migration status from key and my, phonehome is not needed.

packages/ns-api/files/ns.backup
```diff
@@ -169,21 +169,26 @@ elif cmd == 'call':
...
-                print(json.dumps({'message': 'success'}))
-            except subprocess.CalledProcessError as error:
-                print(json.dumps(utils.generic_error(f'remote upload failed')))
-            except RuntimeError as error:
-                print(json.dumps(utils.generic_error(error.args[0])))
+            # Refuse the call before running sysupgrade/uploading, and emit
+            # valid JSON so the HTTP API wraps it as a 422 ValidationError
+            # the UI can render (the previous form printed a Python dict
+            # repr, which was silently dropped upstream and caused the run
+            # modal to stay open after a successful upload).
```
Tbaile: AI generated a way verbose comment to address a simple validation error, that's the case for all the changes, we should keep it simple

packages/ns-plug/files/migrate-to-my
```diff
@@ -0,0 +1,109 @@
...
+commit ns-plug
+EOI
+
+# Repoint vmalert at the native Mimir endpoint right away. vmalert reloads on
+# ns-plug changes, but migrate-to-my does not trigger reload_config, so force a
+# targeted reload here: the running vmalert otherwise keeps its boot-time
+# (legacy) creds and POSTs to /proxy/alerts until the next reload. Reloading
+# now makes the alert path flip together with the credential rotation. Safe
+# no-op if victoria-metrics is not installed.
+/etc/init.d/vmalert reload 2>/dev/null || true
```
Tbaile, line 90: Restart will be better fitted here, leave reload for the service triggers through config, comment is too verbone and returns a (now with 8.8) wrong affirmations, victoria-metrics is always installed

packages/ns-plug/files/migrate-to-my
```diff
@@ -0,0 +1,109 @@
...
+
+# The enterprise feeds hold the credentials inside their URLs, written at
+# registration: rotating the pair here would leave them authenticating as the
+# legacy system, which works only until the old my goes away. Re-run just the
+# feed hooks -- the other register hooks act on a first registration.
+for feed_setup in /usr/sbin/ts-ip /usr/sbin/ts-dns /usr/sbin/distfeed-setup; do
+    [ -x "$feed_setup" ] && "$feed_setup" || true
+done
```
Tbaile, line 98: `ts-ip`, `ts-dns` and `distfeed` might not work with the new credentials. They still point to old my to check, we need to plan the migration for them.

packages/ns-api/files/ns.dedalo
```diff
@@ -45,24 +46,104 @@ def setup(u):
 def login(args):
     u = EUci()
     try:
-        p = subprocess.run(['curl', '-L', '--url', f'https://{args["host"]}/api/login', '--header', 'Content-Type: application/json', '--data-binary', json.dumps(args)], check=True, capture_output=True, text=True)
+        p = subprocess.run(['curl', '-L', '-m', '15', '--connect-timeout', '5', '--url', f'https://{args["host"]}/api/login', '--header', 'Content-Type: application/json', '--data-binary', json.dumps(args)], check=True, capture_output=True, text=True)
```
Tbaile, line 49: Subprocess will leak credentials, switch to python own requests package. Will likely handle most of the work you need to do to adjust the response

packages/ns-api/files/ns.dedalo
```diff
@@ -45,24 +46,104 @@ def setup(u):
...
+
+def oidc_start(args):
+    host = args.get("host") or "my.nethspot.com"
+    u = EUci()
+    unit_name = u.get("dedalo", "config", "unit_name", default="")
+    if not unit_name:
+        with open('/proc/sys/kernel/hostname', 'r') as fp:
+            unit_name = fp.read().strip()
+    try:
+        p = subprocess.run(['curl', '-s', '-L', '-m', '15', '--connect-timeout', '5', '-X', 'POST', '-w', '\n%{http_code}', '--url', f'https://{host}/api/auth/oidc/device/start', '--header', 'Content-Type: application/json', '--data-binary', json.dumps({"unit_name": unit_name})], check=True, capture_output=True, text=True)
```
Tbaile, line 90: Subprocess will leak credentials, switch to python own requests package. Will likely handle most of the work you need to do to adjust the response

packages/ns-api/files/ns.dedalo
```diff
@@ -45,24 +46,104 @@ def setup(u):
...
+def oidc_poll():
+    u = EUci()
+    try:
+        with open(pairing_file, 'r') as fp:
+            pairing = json.load(fp)
+    except:
+        return utils.generic_error("no_pairing_in_progress")
+    host = pairing["host"]
+    try:
+        p = subprocess.run(['curl', '-s', '-L', '-m', '15', '--connect-timeout', '5', '--url', f'https://{host}/api/auth/oidc/device/poll', '--header', 'Content-Type: application/json', '--data-binary', json.dumps({"device_code": pairing["device_code"]})], check=True, capture_output=True, text=True)
```
Tbaile, line 125: Subprocess will leak credentials, switch to python own requests package. Will likely handle most of the work you need to do to adjust the response

packages/ns-api/files/ns.dedalo
```diff
@@ -131,7 +212,7 @@ def list_parents():
     u = EUci()
     try:
         api_url = u.get("dedalo", "config", "api_url")
-        p = subprocess.run(['curl', '-L', '-s', '--url', f'{api_url}/hotspots', '--header', f"Token: {_get_token()}"], capture_output=True, text=True)
+        p = subprocess.run(['curl', '-L', '-s', '-m', '15', '--connect-timeout', '5', '--url', f'{api_url}/hotspots', '--header', f"Token: {_get_token()}"], capture_output=True, text=True)
```
Tbaile, line 215: Subprocess will leak credentials, switch to python own requests package. Will likely handle most of the work you need to do to adjust the response

packages/ns-plug/files/register
```diff
@@ -42,11 +47,25 @@ case "$type" in
...
         ;;
     enterprise)
-        url="https://my.nethesis.it/api/"
+        url="https://my-proxy-prod.onrender.com/backend/api/"
 
-        system_id=$(curl -s -m $timeout --retry 3 -L \
+        register_resp=$(curl -s -m $timeout --retry 3 -L -w '\n%{http_code}' \
         -H "Content-Type: application/json" -H "Accept: application/json" \
-        -d '{"secret": "'$secret'"}' "${url}systems/info" | jq -r ".uuid" 2>/dev/null)
+        -d '{"system_secret": "'$secret'"}' "${url}systems/register")
```
Tbaile, line 49: Since we cannot use python here, we can put the payload of curl inside the stdin, this way we don't show at `ps` the credentials

packages/ns-plug/files/remote-backup
Tbaile, line 1: Here the `curl` has the same issue, it can leak credentials, not sure we can do anything about it since it ends up in the `--user` flag.

packages/ns-plug/files/send-backup
```diff
@@ -45,6 +50,12 @@ if [ -z "$SYSTEM_ID" ] || [ -z "$SYSTEM_SECRET" ]; then
     exit 0
 fi
 
+# remote-backup handles the enterprise/community branching and calls
+# migrate-to-my when needed; this script just prepares the payload and
+# delegates the upload. An enterprise unit still waiting on the
+# migration surfaces its error through remote-backup, which is caught
+# by set -e above.
+
```
Tbaile: not needed comment...?

packages/ns-plug/files/send-heartbeat
```diff
@@ -5,26 +5,50 @@
...
     # System ID or system secret not found, configure ns-plug
     exit 0
 fi
 
-/usr/bin/curl -m 180 --retry 3 -L -s \
-   --header "Authorization: token $SYSTEM_SECRET"  --header "Content-Type: application/json" --header "Accept: application/json"  \
-   --data-raw '{"lk": "'$SYSTEM_ID'"}' "$URL" >/dev/null
+case "$TYPE" in
+    enterprise)
+        /usr/sbin/migrate-to-my
```
Tbaile: MIgration should happen at first boot after the update, not like this, same as the backup process

packages/ns-plug/files/send-heartbeat
```diff
@@ -5,26 +5,50 @@
...
+            "$COLLECT_URL/heartbeat" 2>&1) || \
+            logger -t send-heartbeat "heartbeat send failed: $err"
+        ;;
+    community)
+        URL=$(uci -q get ns-plug.config.alerts_url)"heartbeats/store"
+        /usr/bin/curl -m 180 --retry 3 -L -s \
+            --header "Authorization: token $SYSTEM_SECRET" \
+            --header "Content-Type: application/json" \
+            --header "Accept: application/json" \
+            --data-raw '{"lk": "'$SYSTEM_ID'"}' "$URL" >/dev/null
```
Tbaile: This was already here, but we have leaking credentials here too

packages/ns-plug/files/unregister
```diff
@@ -11,14 +11,28 @@
...
+# and community has no freekey equivalent on dartagnan.
+if [ "$TYPE" = "enterprise" ]; then
+    LEGACY_ID=$(uci -q get ns-plug.config.legacy_system_id)
+    LEGACY_SECRET=$(uci -q get ns-plug.config.legacy_secret)
+    if [ -n "$LEGACY_ID" ] && [ -n "$LEGACY_SECRET" ]; then
+        curl -s -m 180 --retry 3 -L \
+            -H "Content-type: application/json" -H "Accept: application/json" \
+            -d "{\"lk\":\"$LEGACY_ID\",\"secret\":\"$LEGACY_SECRET\"}" \
+            https://my-proxy-prod.onrender.com/api/Utils/freekey >/dev/null
+    fi
```
Tbaile, line 29: Leaking credentials

packages/ns-plug/files/subscription-info
```diff
@@ -6,29 +6,68 @@
...
 
 type=$(uci -q get ns-plug.config.type)
 secret=$(uci -q get ns-plug.config.secret)
-url=$(uci -q get ns-plug.config.api_url | sed 's/\/$//')
 
 if [ "$type" = "enterprise" ]; then
-    curl -f -s -m $timeout --retry-delay 1 --retry 2 -L \
-        -H "Content-Type: application/json" -H "Accept: application/json" \
-        -d '{"secret": "'$secret'"}' "$url/systems/info"
+    /usr/sbin/migrate-to-my
```
Tbaile: Migration here has the same issues as I've noted above

packages/ns-plug/files/subscription-info
```diff
@@ -6,29 +6,68 @@
...
+        exit 1
+    fi
+
+    # /info returns {code, message, data: {system_id, registered,
+    # registered_at, suspended, organization: {name, ...}, ...}}.
+    # Translate to the legacy shape the UI/info handler parses.
+    resp=$(/usr/bin/curl -f -s -m $timeout --retry-delay 1 --retry 2 -L \
+        -H "Accept: application/json" \
+        --user "$system_id:$secret" \
+        "$collect_url/info") || exit $?
```
Tbaile: user leaks here too

packages/ns-plug/files/send-inventory
Tbaile, line 1: This file has the same issues, the migration is fragile and credentials leak

packages/ns-plug/files/migrate-to-my
```diff
@@ -0,0 +1,109 @@
...
+fi
+
+# Fetch the mapped my credentials via the translation proxy.
+resp=$(/usr/bin/curl --silent --location-trusted --fail-with-body \
+    --max-time 30 --retry 2 \
+    --user "$SYSTEM_ID:$SYSTEM_SECRET" \
+    https://my-proxy-prod.onrender.com/proxy/credentials 2>/dev/null) || {
+    logger -t migrate-to-my "credential fetch failed; will retry on next run"
+    exit 0
+}
```
Tbaile, line 63: Credentials leak

Tbaile (changes requested):
This review is harsh due to just the many credentials leak that will happen in case the curl hangs for some reason since they get printed directly into `ps`. Sadly the system is taunted with many of them. 
One of the other major issues is the migration to the new my that gets called whenever we run the backups, heartbeat or inventory.
I can't figure another way to implement this without breaking the old implementation, a `init.d` would be nice but if it doesn't run we encounter issues with all the scripts, it's not acceptable.
We could've got the chance to rewrite many of the components with python to streamline and avoid bash directly, sadly this won't be the case and I don't think right now is possible.

I'll try and put my head around it if I find some solution to the migration, keep it like this for the moment.

packages/ns-api/files/ns.dedalo
gsanchietti: Alternative: read the data from a file or stdin

packages/victoria-metrics/files/vmalert.initd
```diff
@@ -19,25 +19,37 @@ start_service() {
...
         config_get system_secret config secret ""
         config_get system_type config type ""
+        config_get migrated config migrated ""
+        config_get collect_url config collect_url ""
     }
 
+    notifier_url=""
     if [ "$system_type" = "enterprise" ] && [ -n "$system_id" ] && [ -n "$system_secret" ]; then
-        notifier_url="https://my.nethesis.it/proxy/alerts"
+        if [ "$migrated" = "1" ] && [ -n "$collect_url" ]; then
```
Tbaile: to consider subscriptions too, migrated might not needed

packages/ns-threat_shield/files/ts-ip
```diff
@@ -33,8 +33,17 @@ TYPE=$(uci -q get ns-plug.config.type)
...
+    # rotation keeps fetching as the old pair. Enforce one entry, the current
+    # one: compare the whole set, and del_list per entry (a multi-line value
+    # matches nothing).
+    allow_current="https://$SYSTEM_ID:$SYSTEM_SECRET@bl.nethesis.it/plain/$TYPE/nethesis-blacklists/whitelist.global"
+    allow_have=$(uci -q get banip.global.ban_allowurl | tr " " "\n" | grep bl.nethesis.it)
+    if [ "$allow_have" != "$allow_current" ]; then
+        for allow_stale in $allow_have; do
+            uci del_list banip.global.ban_allowurl="$allow_stale"
+        done
+        uci add_list banip.global.ban_allowurl="$allow_current"
```
Tbaile: To try multiple add to allow_url

packages/ns-plug/files/unregister
```diff
@@ -11,14 +11,28 @@
...
-URL=$(uci -q get ns-plug.config.api_url | sed 's/\/$//')
+TYPE=$(uci -q get ns-plug.config.type)
 
 if [ -z "$SYSTEM_ID" ] || [ -z "$SYSTEM_SECRET" ]; then
     # System ID and System secret not found, configure ns-plug to enable it
     exit 0
 fi
 
-curl -s -m 180 --retry 3 -L -H "Content-type: application/json" -H "Accept: application/json" -d "{\"lk\": \"$SYSTEM_ID\", \"secret\": \"$SYSTEM_SECRET\"}" "$URL/Utils/freekey" >/dev/null
+# Release the legacy slot on my-old for migrated enterprise units —
```
Tbaile: to be verified

packages/ns-plug/files/subscription-info
Tbaile: To be moved to postinst

PR #1620, opened 2026-04-29
chore: add more info to agents.md

gsanchietti:
Add more info on external repos and
build info


Tbaile (approved):
Good for me, check the skill activation prompt, if you think is ok, squash merge right away

PR #1621, opened 2026-04-29
feat(snort): add max_inspect_bytes nftables rule for large file handling

### LINKED ISSUE #1624: IPS breaks file transfers larger than 4GB
**Description**

When IPS is enabled on NethSecurity 8, downloads or uploads of files larger than 4GB fail.

The issue seems related to Snort/IPS inspection, because the same transfers complete successfully when IPS is disabled or when bypass rules are configured.

**Steps to reproduce**

1. Enable IPS on NethSecurity 8.
2. Configure IPS with an active ruleset policy.
3. Start a download or upload of a file larger than 4GB.
4. Wait until the transfer reaches 4GB.

**Expected behavior**

The file transfer should complete successfully, regardless of the file size, while IPS is enabled.

**Actual behavior**

The transfer stops or fails when it reaches exactly 4GB.

**Workaround**

The issue can be avoided by disabling IPS

**Components**
NethSecurity 8.7.2

gsanchietti:
## Summary

Add configurable nftables rule that skips snort inspection for connections that have already transferred N bytes, fixing large file download performance issues.

Fixes: https://github.com/NethServer/nethsecurity/issues/1624

## Details

Implements the performance improvement discussed in the team thread:
- Connections transferring >1MB bypass snort inspection (default)
- Reduces performance impact on large downloads while catching attacks in early-stage traffic
- Fully configurable: 0 = unlimited inspection, any other value = byte limit

## Changes

**snort3 package:**
- Added `max_inspect_bytes` UCI option to snort.nfq (range 0–2147483648, default 1048576)
- Generated nftables rule: `ct bytes > N accept` inserted before bypass-IP rules
- Rule is conditionally omitted when max_inspect_bytes=0
- Updated package README with documentation

**ns-api package:**
- Exposed `max_inspect_bytes` in ns.snort API (`settings`, `save-settings`, `list`)
- Updated API README with new field documentation
- Proper validation: accepts integer ≥ 0; rejects negative/invalid values

## How to test

1. SSH to a device with snort enabled
2. Check default nftables rule:
   ```bash
   nft list table inet snort
   # Should show: ct bytes > 1048576 accept
   ```
3. Toggle to unlimited inspection:
   ```bash
   uci set snort.nfq.max_inspect_bytes=0
   snort-mgr resetup
   nft list table inet snort
   # Should NOT show ct bytes rule
   ```
4. Set custom value via API:
   ```bash
   echo '{"enabled": true, "ns_policy": "connectivity", "oinkcode": "", "home_net": ["192.168.100.0/24"], "max_inspect_bytes": 0 }' | /usr/libexec/rpcd/ns.snort call save-settings
   ```
5. Verify via live device: large file downloads complete without stalls

## Testing notes

Check the following:
- Default value (1MB) generates correct nftables rule
- Setting to 0 omits rule entirely
- API save-settings accepts and persists custom values
- nftables reload applies changes immediately

### DIFF
--- a/packages/ns-api/README.md
+++ b/packages/ns-api/README.md
@@ -8037,16 +8037,18 @@ Response example:
   "enabled": true,
   "ns_policy": "balanced",
   "oinkcode": "123456789",
-  "home_net": ["192.168.1.0/24", "10.0.0.0/8"]
+  "home_net": ["192.168.1.0/24", "10.0.0.0/8"],
+  "max_inspect_bytes": 1048576
 }
 ```
 
 ### save-settings
 
 Set `snort` configuration. The `home_net` field is a list of IPv4 CIDRs representing the protected networks.
+The optional `max_inspect_bytes` field sets how many bytes per connection are forwarded to snort for inspection; use `0` for unlimited (default: `1048576`).
 
 ```bash
-api-cli ns.snort save-settings --data '{"enabled": true, "ns_policy": "balanced", "oinkcode": "123456789", "home_net": ["192.168.1.0/24"]}'
+api-cli ns.snort save-settings --data '{"enabled": true, "ns_policy": "balanced", "oinkcode": "123456789", "home_net": ["192.168.1.0/24"], "max_inspect_bytes": 1048576}'
 ```
 
 ### check-oinkcode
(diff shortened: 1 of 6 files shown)


cotosso:
I tested the PR using the image `nethsecurity-8.7.2-snort-performance+a26b08a84-x86-64-generic-squashfs-combined-efi.img.gz`.

Everything went as expected, all the modifications made by both uci and via API were correctly reflected in the nft snort table.
I was finally able to download a 5 GB test file with no hassle.

filippocarletti (approved):

PR #1603, opened 2026-04-21
refactor: using custom script to configure netifyd

Tbaile:
Supporting bypasses for the new nfqueue ingesting method.

Closes: https://github.com/NethServer/nethsecurity/issues/1628
Closes: https://github.com/NethServer/nethsecurity/issues/1627

PR #1626, opened 2026-04-30
feat: implement netifyd bypass

gsanchietti:
Some hosts or networks may cause so much traffic
that netifyd crashes for OOM


gsanchietti:
Let's go with implementation #1603

PR #1632, opened 2026-04-30
feat(openvpn): add tun_mtu and mssfix configuration management for instances

### LINKED ISSUE #1192: Add mtu and mssfix configuration options to advanced settings in OpenVPN (RW and tunnels)
**Description:**

It would be beneficial to enhance the OpenVPN GUI by incorporating configuration options for `tun-mtu` and `mssfix` within the Advanced Settings section. These parameters are critical not only for performance optimization but also for ensuring proper functionality of certain services that may break when the MTU is too high.

**Justification:**

- **MTU Adjustments:** The `tun-mtu` parameter allows users to specify the Maximum Transmission Unit size for the tunnel interface. A value that is too high can lead to packet fragmentation or blocking, which in turn can cause certain services (such as web applications or APIs) to malfunction or become unreachable.

- **MSSFix Configuration:** The `mssfix` option helps in setting the Maximum Segment Size for TCP connections to avoid IP fragmentation. It is particularly useful when the underlying network has smaller MTU constraints.

**Proposed Implementation:**

- Add input fields for `tun-mtu` and `mssfix` in the Advanced Settings tab of the OpenVPN GUI.
- Include contextual help explaining recommended values and scenarios where adjustments may be necessary.

**References:**
https://docs.nethsecurity.org/en/latest/openvpn_roadwarrior.html#mtu-issue-and-packet-fragmentation

### LINKED ISSUE #1427: OpenVPN: Revert tun_mtu and mssfix values to OpenVPN 2.5 defaults to avoid MTU-related traffic issues
**Description**
Starting with OpenVPN 2.6, the default value of _mssfix_ has changed compared to previous versions (1492 instead of 1450). This modification has caused issues in several installations: when the actual line MTU is slightly below 1500, VPN traffic may fail to pass correctly or become unstable due to packet fragmentation problems.

**Proposed solution**

Set the following defaults in the OpenVPN configuration:

```
tun_mtu = 1500  
mssfix = 1450
```

This way we align the behavior with OpenVPN 2.5 defaults, which were stable and did not generate user-visible issues. Lowering mssfix to 1450 is transparent to end users and effectively prevents fragmentation-related connectivity problems.

m-dilorenzi:
This pull request adds support for managing and defaulting the `tun_mtu` and `mssfix` parameters for OpenVPN Roadwarrior and server/client tunnels across the codebase. These parameters are now consistently set, exported, and included in configuration and API responses, ensuring improved interoperability and reducing MTU-related connection issues. The changes also introduce a new script to backfill these defaults for existing instances.

**OpenVPN MTU and MSSFix Support**

* Default values for `tun_mtu` (1500) and `mssfix` (1450) are now set when creating or importing OpenVPN instances, both for server/client tunnels and RoadWarrior Servers.
* These parameters are included in exported client/server configurations and are now part of the API responses and example payloads for tunnel management endpoints
* The OpenVPN user configuration file now includes `tun-mtu` and `mssfix` directives, using either the configured or default values.
* The `set_configuration` logic now allows updating `tun_mtu` and `mssfix` values for an instance if changed. 
* A new initialization script (`99_ns-openvpn-mtu-mssfix-defaults`) is added to ensure all existing OpenVPN instances have `tun_mtu` and `mssfix` set to defaults if missing. 

Closes: https://github.com/NethServer/nethsecurity/issues/1192, https://github.com/NethServer/nethsecurity/issues/1427

### DIFF
--- a/packages/ns-api/files/ns.ovpnrw
+++ b/packages/ns-api/files/ns.ovpnrw
@@ -589,10 +589,9 @@ def set_configuration(args):
     else:
         return utils.validation_error("ns_auth_mode", "invalid_auth_mode", args["ns_auth_mode"])
 
-    if args["tun_mtu"] and args["tun_mtu"] != u.get("openvpn", ovpninstance, "tun_mtu", default=''):
-        u.set("openvpn", ovpninstance, "tun_mtu", args.get("tun_mtu"))
-    if args["mssfix"] and args["mssfix"] != u.get("openvpn", ovpninstance, "mssfix", default=''):
-        u.set("openvpn", ovpninstance, "mssfix", args.get("mssfix"))
+    # MTU and MSSFix settings
+    u.set("openvpn", ovpninstance, "tun_mtu", args.get("tun_mtu"))
+    u.set("openvpn", ovpninstance, "mssfix", args.get("mssfix"))
 
     u.save("openvpn")
 
--- a/packages/ns-api/files/ns.ovpnrw
+++ b/packages/ns-api/files/ns.ovpnrw
@@ -589,6 +589,10 @@ def set_configuration(args):
     else:
         return utils.validation_error("ns_auth_mode", "invalid_auth_mode", args["ns_auth_mode"])
 
+    if args["tun_mtu"] and args["tun_mtu"] != u.get("openvpn", ovpninstance, "tun_mtu", default=''):
+        u.set("openvpn", ovpninstance, "tun_mtu", args.get("tun_mtu"))
+    if args["mssfix"] and args["mssfix"] != u.get("openvpn", ovpninstance, "mssfix", default=''):
+        u.set("openvpn", ovpninstance, "mssfix", args.get("mssfix"))
 
     u.save("openvpn")
 
@@ -1232,7 +1236,9 @@ if cmd == 'list':
             "ns_dhcp_options": [],
             "ns_public_ip": ["1.2.3.4"],
             "ns_user_db": "main",
-            "ifconfig_pool": ["10.166.54.4", "10.166.54.20"]
+            "ifconfig_pool": ["10.166.54.4", "10.166.54.20"],
+            "tun_mtu": "1500",
+            "mssfix": "1450",
             },
         "remove-instance": {"instance": "roadwarrior1"},
         "import-users": {"instance": "roadwarrior1"},
--- a/packages/ns-api/files/ns.ovpnrw
+++ b/packages/ns-api/files/ns.ovpnrw
@@ -333,6 +333,8 @@ def add_instance():
     u.set("openvpn", instance, 'tls_version_min','1.2')
     u.set("openvpn", instance, 'ns_auth_mode', 'certificate')
     u.set('openvpn', instance, 'ns_tag', ["automated"])
+    u.set('openvpn', instance, 'tun_mtu', '1500')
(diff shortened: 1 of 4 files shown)


packages/ns-api/files/ns.ovpnrw
```diff
@@ -589,6 +589,10 @@ def set_configuration(args):
     else:
         return utils.validation_error("ns_auth_mode", "invalid_auth_mode", args["ns_auth_mode"])
 
+    if args["tun_mtu"] and args["tun_mtu"] != u.get("openvpn", ovpninstance, "tun_mtu", default=''):
+        u.set("openvpn", ovpninstance, "tun_mtu", args.get("tun_mtu"))
+    if args["mssfix"] and args["mssfix"] != u.get("openvpn", ovpninstance, "mssfix", default=''):
+        u.set("openvpn", ovpninstance, "mssfix", args.get("mssfix"))
```
Tbaile: You can avoid the comparison with the `u.get`s, if you set a value that is the same as the old one UCI translates this in a NO-OP. This applies only to integers and strings, not lists.

packages/ns-api/files/ns.ovpntunnel
```diff
@@ -391,6 +399,10 @@ def list_tunnels():
                 "remote_host": vpn.get("remote", ""),
                 "remote_network": remote
             }
+            if vpn.get("tun_mtu"):
+                client["tun_mtu"] = vpn.get("tun_mtu")
+            if vpn.get("mssfix"):
+                client["mssfix"] = vpn.get("mssfix")
```
Tbaile: Careful, in the UI these fields are mandatory per type definition: https://github.com/NethServer/nethsecurity-ui/pull/753/changes#diff-164c0ee5687576db4d4524d909216099b5c997e0bd169bc5bdc0d1a3b64ad693R49-R50

This applies to both client and server definitions, you could use
```
client["tun_mtu"] = vpn.get("tun_mtu", "default-value or something")
```

Tbaile (changes requested):

packages/ns-api/files/ns.ovpntunnel
```diff
@@ -191,6 +191,10 @@ def import_client(tunnel):
 
     # Add management socket
     u.set("openvpn", iname, 'management', f'/var/run/openvpn_{iname}.socket unix')
+    
+    # set tun_mtu and mss_fix to default values if not provided by client config
+    u.set("openvpn", iname, 'tun_mtu', tunnel.get('TunMtu', '') or '1500')
+    u.set("openvpn", iname, 'mssfix', tunnel.get('Mssfix', '') or '1450')
```
Tbaile, line 197:
```suggestion
    u.set("openvpn", iname, 'tun_mtu', tunnel.get('TunMtu', '1500'))
    u.set("openvpn", iname, 'mssfix', tunnel.get('Mssfix', '1450'))
```

Why not this? Is there something I'm missing?

m-dilorenzi: Prevents assigning the empty string ('') to the two properties. If the values arrive as MssFix='' and/or TunMtu='', the first approach would set them to 1450 and 1500 respectively; the second approach would assign the empty string instead, because the key exists and its value (not the default) is used.

Tbaile (approved):

PR #1601, opened 2026-04-20
Mimir integration

gsanchietti:
:warning:  The PR has been tested and should work in common scenarios.
Waiting the release of new my.nethesis.it before releasing

---

This pull request new features inside monitoring and alerting system:

- it adds a script (`ns-plug-alert`) to send alerts to the new implementation of my.nethesis.it using alertmanager format
- it fixes netdata configuration update

To test, setup the URL and credentials:
```
uci set ns-plug.config.my_url=https://qa.my.nethesis.it
uci set ns-plug.config.my_system_key=NETH-F5D2-5E69-A174-45A9-B1AB-2BB9-xxxx-xxxx
uci set ns-plug.config.my_system_secret=my_f620a0c63a99161021aa.21af8bffebd96356a6e7bbdee67321ba4bexxxxx
uci commit
```

Then:
```
ns-plug-alert fire --alertname DiskSpaceCritical --severity critical   --labels service=storage mountpoint=/mnt/data   --annotations     "summary_en=Disk space critical"     "summary_it=Spazio disco critico"     "description_en=Disk usage above 90% on /mnt/data"     "description_it=Utilizzo disco sopra 90% su /mnt/data"

ns-plug-alert resolve --alertname DiskSpaceCritical --severity critical   --labels service=storage mountpoint=/mnt/data   --annotations     "summary_en=Disk space critical"     "summary_it=Spazio disco critico"     "description_en=Disk usage above 90% on /mnt/data"     "description_it=Utilizzo disco sopra 90% su /mnt/data"
{"status": "success", "message": "Alert 'DiskSpaceCritical' resolved"}
```

PR #1633, opened 2026-04-30
Alerts and charts using Victoria Metrics

### LINKED ISSUE #1656: Monitoring: adding long time storage for metrics
Firewall needs a long term storage for metrics, at the moment Netdata does the job but it has the following limitations:
- only 6 hours
- very old build
- hard to customize

We landed on [Victoria Metrics](https://victoriametrics.com/) due to the nice space usage and easy replacement for Prometheus, which does the same as Netdata.
However a tool that scrapes the data from the firewall and send it to Victoria is needed, that's where [Telegraf](https://github.com/influxdata/telegraf) comes in.
From Telegraf the data to be sent to the Victoria Metrics is the following:
- generic machine info (CPU, MEM, etc...)
- MWAN status
- services status
- pings

Victoria can through `vmalert` mimic what AlertManager does, which is in firewall's case:
- notify downtimes
- notify alerts

Alerts is kept in storage by Victoria Metrics itself.

gsanchietti:
This pull request migrates network monitoring and reporting functionality from Netdata to Telegraf and Victoria Metrics,
It raises alerts using Mimir as alertmanager for Victoria Metrics.
It keeps netdata with its own alerting for backward compatibility.

Main changes:
- ns.dashboard and ns.report APIs now uses Victoria Metrics statistics
- add a new Metrics page where most relevant metrics are shown (this should replace Netdata UI)
- ping latency monitor had been ported from Netdata to Telegraf: ns.telegraf replaces ns.netdata, output and API name for the UI is unchanged
- implement mwan alerts using a new Telegraf plugin
- implement service alerts using a new Telegraf plugin: only most relevant services are monitored
- port storage and backup alerts to Telegraf plugin
- send alerts to current my.nethesis.it and my.nethserver.com using a local proxy that mimic Alertmanager behavior
- remove Netdata
- add Mimir support for next-gen my.nethesis.it

Replaces #1601 
Reference: #1638


Tbaile (approved):
I really don't like that we use `vm_query` for bridging requests over to the API and format the data, we could allow later on directly the UI to do the requests directly to the victoria metrics instance. This will need an edit to the `ns-api-server`, which I'm not that fond of, leave the merge for the moment, going to check for issues on the UI and merge everything all together. 

Check my latest commits for a few changes, nothing big but another couple of eyes never hurt.

gsanchietti:
> I really don't like that we use `vm_query` for bridging requests over to the API and format the data, we could allow later on directly the UI to do the requests directly to the victoria metrics instance

Me neither but I think is good enough for most cases. Let's keep it as is, we are going to improve the implementation if it's too heavy on real machines.

> Check my latest commits for a few changes, nothing big but another couple of eyes never hurt.

It's good for me.

Please go with the merge when the UI part is good for you (it's a bit ugly right now, but it works and for me we can merge as is).

Issue #1660, opened 2026-05-12
Slow IPsec page loading with high number of tunnels


federicoballarini:
Following recent configuration changes and updates, a significant increase in latency has been observed while trying to load IPsec management page, particularly in environments managing over 100 tunnels. Beyond the general slowdown, this issue results in critical UI malfunctions:
- The "Apply Changes" button frequently fails to render.
- In some instances, the interface displays inconsistent or stale data when the button eventually appears.

**Steps to reproduce**

- Create and connect 50+ IPsec tunnels
- Try to load the IPsec page and edit some settings

**Expected behavior**

The IPsec management page should load efficiently and remain responsive.

**Actual behavior**

The IPsec managment page is slow and need more than 10 seconds to load.

**Components**

NethSecurity 8.7.2

**See also**

Nethesis Helpdesk ticket #201753

gsanchietti:
Verified by Federico

PR #1662, opened 2026-05-13
fix(ns-api): ipsectunnels, call swanctl once

gsanchietti:
The list-tunnels endpoint was calling swanctl
command multiple times, one for each existing tunnel.

If the machine had a list of many tunnels, the page was really slow.

Issue #1660


Tbaile (approved):
🤷

PR #1677, opened 2026-05-15
chore: trying to build pyton3 from commits instead of versions

### LINKED ISSUE #1617: Add "DON'T TRACK" action to firewall rules
**Description**
Add a new firewall rule action called **"DON'T TRACK"** alongside the existing actions (ACCEPT, REJECT, DROP).
This is needed to support use cases where connection tracking (conntrack) must be bypassed, improving performance and enabling proper handling of specific traffic types (e.g. high-throughput flows, stateless protocols, or asymmetric routing scenarios).
The purpose is to expose, at UI level, a feature already supported by underlying netfilter/nftables.

**Proposed solution**
Introduce a new selectable action **DON'T TRACK** in firewall rules.
When selected, the rule should apply the equivalent of a `notrack` (raw table) behavior, ensuring packets matching the rule are excluded from connection tracking.
The target option in firewall config must be `NOTRACK`.

The option should be available in the same place where ACCEPT/REJECT/DROP are currently defined, maintaining UI consistency.

**Additional context**
This functionality is already supported at lower levels (netfilter/nftables), but currently not exposed in NethSecurity UI. Making it available improves flexibility for advanced networking scenarios without requiring manual configuration.

Tbaile:
Pointing the last compiling version to the last commit of `nethsecurity-8.8`.

### DIFF
--- a/packages/python3-nethsec/Makefile
+++ b/packages/python3-nethsec/Makefile
@@ -13,11 +13,15 @@ PKG_RELEASE:=1
 PKG_MAINTAINER:=Giacomo Sanchietti <giacomo.sanchietti@nethesis.it>
 PKG_LICENSE:=GPL-3.0-only
 
-PKG_SOURCE:=python3-nethsec-$(PKG_VERSION).tar.gz
-PKG_SOURCE_URL:=https://codeload.github.com/nethserver/python3-nethsec/tar.gz/$(PKG_VERSION)?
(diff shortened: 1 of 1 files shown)


gsanchietti (approved):
🤷 I still prefer to use versions
Otherwise we can directly merge the python3 package inside the main repo

PR #1678, opened 2026-05-15
refactor: using http sink for flows ingestion

### LINKED ISSUE #1681: flows: move from socket listener to http
Seems that sometimes socketing in go fails to get the right process at first, attaching itself to a file descriptor empty (thus hanging there forever). Turn the ns-flows package like the ns-stats one, a fiber server with http server available that gets sent data from netifyd.

Tbaile:
Also updated the binary to support GO 1.26.2


gsanchietti (changes requested):
Add a uci default that cleanups old configuration files

Tbaile:
Files are tracked by apk/opkg, not defined as configs, while updating they get pruned automatically since not needed

gsanchietti (approved):

PR #1680, opened 2026-05-15
feat: enable persistent var (EXPERIMENTAL)

gsanchietti:
EXPERIMENTAL: test how the image behaves

Upstream commit: https://github.com/openwrt/openwrt/commit/57807f50ded6cf0996284a850084183af13d5894


gsanchietti:
The persistent var could be useful to retain the following files:
- `/var/log/lastlog` is always empty
- `/var/log/messages` is already stored also in `/mnt/data`
- `/var/log/wtmp` could be useful, but we do not have the `last` command to parse it, also similar info is already present inside the messages

But has many drawbacks:
- upstream implementation does not support mounting a different partition under /var
- it stores also many files in /vat/etc that are recreated by applications upon starts, such files should be temporary
- the messages file can get bigger and fill all the root
- victoria metrics are stored inside /var/lib/victoria-metrics-data and can get bigger and fill the root
- despite of the linked commit, having a persistent var does not preserve dhcp leases across reboots because they are saved by default on a temp directory (`dhcp.ns_dnsmasq.leasefile='/tmp/dhcp.leases'`)

In the end, I think that persistent may cause many problems on long running machines.
My opinion is to ditch the feature and eventually move the DHCP lease to storage, when present.

gsanchietti:
Closing this for now

PR #1683, opened 2026-05-18
build: added back snmpd

Tbaile:
Seems the package got renamed? Was missing from final manifest.
Asking review to be sure we're talking about the same package, there's also the `nossl` version in case (unlikely)


gsanchietti (approved):
Looking the upstream package, ti seems correct

PR #1685, opened 2026-05-19
fix(remote-backup): improve backup name

gsanchietti:
Use a mnemonic name for backup instead of
a random name derived from temporary file.

### DIFF
--- a/packages/ns-plug/files/remote-backup
+++ b/packages/ns-plug/files/remote-backup
@@ -59,9 +59,11 @@ case "$cmd" in
        # proxy, same pattern used by send-heartbeat / send-inventory.
        # To be removed once the migration is complete.
        if [ "$TYPE" = "enterprise" ]; then
+           ext=".$(basename $file | cut -d '.' -f2)"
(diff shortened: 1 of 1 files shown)


packages/ns-plug/files/remote-backup
```diff
@@ -59,9 +59,11 @@ case "$cmd" in
        # proxy, same pattern used by send-heartbeat / send-inventory.
        # To be removed once the migration is complete.
        if [ "$TYPE" = "enterprise" ]; then
+           ext=".$(basename $file | cut -d '.' -f2)"
+           date="-$(date +%Y-%m-%d)"
            curl $curl_args -X POST \
                -H "Content-Type: application/octet-stream" \
-               -H "X-Filename: $(basename "$file")" \
+	       -H "X-Filename: backup$date$ext" \
```
edospadoni: This line is indented with a tab followed by spaces, while the rest of the block uses 15 spaces. Aligning it with the surrounding lines:

```suggestion
               -H "X-Filename: backup$date$ext" \
```

Issue #1694, opened 2026-05-26
Preserve dnsmasq DHCP leases on mounted storage


gsanchietti:
## Description

Persist dnsmasq DHCP leases across reboots when NethSecurity storage is configured on /mnt/data.

When storage is mounted, dnsmasq should use a persistent lease file under `/mnt/data/dnsmasq/dhcp.leases`.

When storage is not mounted, the system must keep using `/tmp/dhcp.leases`.

If storage is later unmounted or enters an error state, dnsmasq must switch back to `/tmp/dhcp.leases` before `/mnt/data` is removed.

The target directory must be created automatically before switching dnsmasq, otherwise the service will fail to start.

This preserves address continuity for clients across reboots, especially on systems with a small DHCP pool.

## Alternative solutions

- Keep using `/tmp/dhcp.leases` and periodically copy it to persistent storage. Rejected because dnsmasq would still restart from a volatile lease file after reboot.
- Update every consumer to read the storage-backed lease file directly. Rejected because the change should stay inside `ns-storage` and keep existing `/tmp/dhcp.leases` consumers compatible.

## References

- Requested flow: keep `/tmp` when storage is absent, switch to storage when mounted, revert to `/tmp` before unmount.
- Current config example: `dhcp.ns_dnsmasq.leasefile='/tmp/dhcp.leases'`

gsanchietti:
Testing image: latest from nethsecurity-8.8 branch

To check dnsmasq config:
```
uci show dhcp.ns_dnsmasq.leasefile
```

**Test case 1**
- Install a clean machine without an extra storage
- Verify the dhcp lease file is saved inside /tmp

**Test case 2**
- After test case 1
- Add a storage
- Verify the dhcp lease file is saved inside /mnt/data

**Test case 3**
- After test case 2
- Remove the storage
- Verify the dhcp lease file is saved inside /tmp

cotosso:
I tested it on VirtualBox with an additional disk for the storage  
(NethSecurity 8.8.0-nethsecurity-8.8.20260529084730.a4fe03716 v25.12.4)


**Test case 1 : OK**
**Test case 2 : OK**

**Test case 3 :**
I can't remove the storage from the UI 

<img width="967" height="559" alt="Image" src="https://github.com/user-attachments/assets/0210e624-7481-4e9d-9fbe-6d463bcc158e" />

this is the command:

```
root@NethSec:~# /usr/libexec/rpcd/ns.storage call remove-storage
[...]

PR #1695, opened 2026-05-26
Persist dnsmasq DHCP leases on mounted storage

gsanchietti:
## Summary

- store dnsmasq DHCP leases on `/mnt/data/dnsmasq/dhcp.leases` when storage is mounted
- keep `/tmp/dhcp.leases` as the compatibility path and fall back to it before storage removal
- restart victoria-metrics around fstab changes so `remove-storage` can release `/mnt/data`

## Related issue

#1694

## How to test

1. Confirm `uci get dhcp.ns_dnsmasq.leasefile` is `/tmp/dhcp.leases` when storage is not configured.
2. Run `add-storage /dev/<data-partition>` and verify the lease path switches to `/mnt/data/dnsmasq/dhcp.leases`, `/tmp/dhcp.leases` becomes a symlink, and the storage lease file is created.
3. Reboot and confirm the lease file on `/mnt/data` is preserved.
4. Run `remove-storage` and verify dnsmasq returns to `/tmp/dhcp.leases` before `/mnt/data` is unmounted.

## Dependencies

None.

### DIFF
--- a/packages/ns-storage/Makefile
+++ b/packages/ns-storage/Makefile
@@ -6,7 +6,7 @@
 include $(TOPDIR)/rules.mk
  
 PKG_NAME:=ns-storage
-PKG_VERSION:=1.0.1
+PKG_VERSION:=1.0.2
 PKG_RELEASE:=1
  
 PKG_BUILD_DIR:=$(BUILD_DIR)/ns-storage-$(PKG_VERSION)
@@ -62,6 +62,7 @@ define Package/ns-storage/install
 	$(INSTALL_BIN) ./files/ns-storage-setup-partition $(1)/usr/libexec
 	$(INSTALL_BIN) ./files/ns-storage-setup-disk $(1)/usr/libexec
 	$(INSTALL_BIN) ./files/ns-storage-has-free-space $(1)/usr/libexec
+	$(INSTALL_BIN) ./files/ns-storage-dhcp-leases $(1)/usr/libexec
 	$(INSTALL_BIN) ./files/remove-storage $(1)/usr/sbin
 	$(INSTALL_BIN) ./files/rotate-messages $(1)/usr/sbin
 	$(INSTALL_BIN) ./files/sync-data $(1)/usr/sbin
--- a/packages/ns-storage/files/add-storage
+++ b/packages/ns-storage/files/add-storage
@@ -48,3 +48,6 @@ crontab -l | grep -q '/usr/sbin/sync-data' || echo '40 1 * * * /usr/sbin/sync-da
 
 # check for existing OpenVPN RW connection records in /var and merge them into storage if needed
 /usr/libexec/ns-openvpn/openvpn-merge-connections-db
+
+# Setup persistent dnsmasq leases file
+/usr/libexec/ns-storage-dhcp-leases
--- a/packages/ns-storage/files/ns-storage-dhcp-leases
+++ b/packages/ns-storage/files/ns-storage-dhcp-leases
@@ -1,0 +1,49 @@
+#!/bin/sh
+#
+# Copyright (C) 2026 Nethesis S.r.l.
+# SPDX-License-Identifier: GPL-2.0-only
+#
+
+# This script is used to move the dnsmasq leases file between the storage and the tmpfs,
+# depending on the storage availability.
+
+TMP_LEASEFILE=/tmp/dhcp.leases
+STORAGE_TARGET=/mnt/data
+STORAGE_DIR=${STORAGE_TARGET}/dnsmasq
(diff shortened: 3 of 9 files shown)


packages/ns-storage/files/10-ns-storage-dhcp-leases
```diff
@@ -0,0 +1,25 @@
...
+
+case "$ACTION" in
+	add)
+		i=0
+		while [ "$i" -lt 10 ] && ! awk '$2 == "/mnt/data" {found=1} END {exit !found}' /proc/mounts; do
+			sleep 1
+			i=$((i + 1))
+		done
+		;;
+	remove)
```
stephdl: If the 10s loop times out and `/mnt/data` never appears, we still fall through to call `ns-storage-dhcp-leases` — that would run in `auto` mode, find no mount, and trigger a pointless dnsmasq restart back to `/tmp`. Should we add an explicit exit after the loop if the mount is not found?

packages/ns-storage/files/10-ns-storage-dhcp-leases
```diff
@@ -0,0 +1,25 @@
...
+		;;
+	*)
```
stephdl: The `remove` case is empty and falls through to the `ns-storage-dhcp-leases` call at the bottom. At that point `/mnt/data` might still be mounted, so `auto` mode could switch dnsmasq back to storage instead of tmp. `remove-storage` already handles the tmp switch explicitly — should we just `exit 0` here to avoid a potential conflict?

packages/ns-storage/files/ns-storage-dhcp-leases
```diff
@@ -0,0 +1,135 @@
...
+}
+
+prepare_storage_leasefile()
+{
+	current="$1"
+
+	mkdir -p "${STORAGE_DIR}"
+	if [ "${current}" = "${TMP_LEASEFILE}" ] && [ -e "${TMP_LEASEFILE}" ] && [ ! -L "${TMP_LEASEFILE}" ]; then
+		if [ ! -e "${STORAGE_LEASEFILE}" ] || [ -s "${TMP_LEASEFILE}" ]; then
+			cp -af "${TMP_LEASEFILE}" "${STORAGE_LEASEFILE}"
```
stephdl: The `[ -s "${TMP_LEASEFILE}" ]` condition means a non-empty `/tmp/dhcp.leases` overwrites whatever is already on the storage leasefile. If the storage file had newer leases from before a reboot (and `/tmp` got a stale or shorter file), we would silently lose them. Would it be safer to copy tmp to storage only when the storage file does not exist yet, and skip the overwrite otherwise?

packages/ns-storage/files/add-storage
```diff
@@ -43,6 +45,8 @@ uci set rsyslog.ns_data.destination="$mdir/log/messages"
 uci commit rsyslog
 /etc/init.d/rsyslog restart
 
+/usr/libexec/ns-storage-dhcp-leases storage
+
 crontab -l | grep -q '/usr/sbin/sync-data' || echo '40 1 * * * /usr/sbin/sync-data' >> /etc/crontabs/root
 /etc/init.d/cron restart
```
stephdl, line 47: The exit code of `ns-storage-dhcp-leases storage` is not checked here — if it fails, `add-storage` continues silently. Also, since `ns-storage-check.init` runs at START=11 and the block hotplug can fire around the same time on boot, two concurrent invocations could race and restart dnsmasq twice. Would a `flock` guard in the script and a `|| exit 1` here be worth it?

gsanchietti: Not a problem if it fails.

Tbaile (commented):
Definitely not fond of a hotplug that writes down uci changes, the alternative should be to have a proc.d that listens on changes of fstab configuration, not sure it's the scope here.

gsanchietti:
I agree @Tbaile: I'm going to remove the hotplug part: It should not be needed 99% of the time.

packages/ns-storage/Makefile
```diff
@@ -54,6 +54,7 @@ endef
 define Package/ns-storage/install
 	$(INSTALL_DIR) $(1)/usr/sbin
 	$(INSTALL_DIR) $(1)/etc/uci-defaults
+	$(INSTALL_DIR) $(1)/etc/hotplug.d/block
```
Tbaile: Little left out, you can merge afterwards

Tbaile (changes requested):

packages/ns-storage/Makefile
gsanchietti: Fixed

PR #1699, opened 2026-05-27
feat: add ns-audit [EXPERIMENTAL]

gsanchietti:
Export an audit report for compliance.

### Generate an audit report
```bash
ubus call ns.audit generate '{}'
```

or via API CLI:
```bash
api-cli ns.audit generate --data '{}'
```

### Expected behavior
- Report files are generated in `/var/run/ns-audit/<report_id>/`
- A `tar.gz` bundle is automatically created containing all artifacts
- The `latest` symlink is automatically created/updated to point to the newest report
- The API response includes a `"bundle"` field with the tar.gz path

### List reports
```bash
ubus call ns.audit list-reports '{}'
```

### Get report summary
```bash
ubus call ns.audit get-summary '{"report_id":"<report_id>"}'
```

### Verify tar.gz contents
```bash
tar -tzf /var/run/ns-audit/latest/audit-report.tar.gz | head -20
```

### Access latest report
The `latest` symlink always points to the most recent report:
```bash
ls -la /var/run/ns-audit/latest
```


gsanchietti:
The idea is good but it needs to be refined.
Closing for now, lets keep the branch for future work.

Issue #1572, opened 2026-03-20
Adblock restarts with empty lists after repeated blocklist/allowlist additions or deletions


cotosso:
### Description

When adding or removing domains in DNS Threat Shield (blocklist and allowlist), each operation is immediately applied and triggers a restart of the adblock service.

If multiple additions or deletions are performed in rapid sequence, the service may restart multiple times before previous executions complete. This can lead to a race condition where adblock ends up running without loading any lists, resulting in zero blocked domains and an inconsistent runtime state.

### Steps to reproduce

1. Go to Threat Shield DNS 
2. Open blocklist or allowlist management
3. Add multiple domains quickly (one after another)
4. Check adblock status

### Expected result

Changes should be staged and applied only after user confirmation, ensuring a single controlled service restart.

Adblock should remain in a consistent "enabled" state with populated blocked domains.

### Actual result

Each domain insertion triggers an immediate restart.

If multiple domains are added rapidly, a race condition may occur, resulting in adblock entering a broken state (adblock_status is running, blocked_domains are 0).

```
root@fw [P]:~# /etc/init.d/adblock status
::: adblock runtime information
  + adblock_status  : running
  + adblock_version : 4.1.5
  + blocked_domains : 0
  + active_sources  : malware_lvl2, yoroi_malware_level1, yoroi_malware_level2, yoroi_susp_level1, yoroi_susp_level2
  + dns_backend     : dnsmasq (-), /tmp/dnsmasq.d
  + run_utils       : download: /usr/libexec/wget-ssl, sort: /usr/libexec/sort-coreutils, awk: /usr/bin/gawk
  + run_ifaces      : trigger: -, report: eth4
  + run_directories : base: /tmp, backup: /tmp/adblock-Backup, report: /tmp/adblock-Report, jail: /tmp
  + run_flags       : backup: ✔, flush: ✘, force: ✔, search: ✘, re-dev+789942c16.20260306165641 v24.10.5
port: ✔, mail: ✘, jail: ✘
  + last_run        : -
  + system          : INTEL Corporation SHARKBAY, NethSecurity 8.7.1

```

Instead of the correct state:

```
root@fw [P]:~# /etc/init.d/adblock status
::: adblock runtime information
  + adblock_status  : enabled
  + adblock_version : 4.1.5
  + blocked_domains : 811445
  + active_sources  : malware_lvl2, yoroi_malware_level1, yoroi_malware_level2, yoroi_susp_level1, yoroi_susp_level2
[...]

PR #1703, opened 2026-05-28
fix(adguard): restore 8.7.2 behavior

gsanchietti:
Changes:
- read only custom feed
- preserve local allowlist and blocklist across image update

Issue #1572


packages/adblock/files/99_adblock_migrate_lists.sh
```diff
@@ -11,6 +11,9 @@ uci -q get adblock.ns_lists >/dev/null 2>&1 && exit 0
 
 uci set adblock.ns_lists=ns_lists
 
+mv /etc/adblock/adblock.whitelist /etc/adblock/adblock.allowlist 2>/dev/null || :
+mv /etc/adblock/adblock.blacklist /etc/adblock/adblock.blocklist 2>/dev/null || :
```
Tbaile, line 16: Let's mark it with a comment so that we remember what are the actual changes we've made to upstream

gsanchietti: done

PR #1704, opened 2026-06-03
ci: updated branches and dev process

Tbaile:
# Proposal changes of dev processes

At the moment the dev process is as the following:
- Changes are applied into `main` that produces a DEV build. This dev build is used for testing only
- To release the packages, we need to bump the packages into the main, wait for a dev build to pass, then publish the packages manually. We lock this way merges to main until we're done publishing.
- To release an image, we tie it to a TAG, which is not the ideal given how OpenWRT acts on builds. Source code should be the truth of the build version and if we ever need to release a new image over the already published one we need to rewrite git history.

Current proposal is to automatically push the packages and images when merged to specified branches:
- `main` doesn't change, it's the dev branch and won't be touched, processes here happen as it always has been.
- `pr` pull request also don't actually change.
- `staging` introducing this branch to allow us to have a staging channel for some partners and/or specified machine we want test the not ready to release packages. This will be updated by merging the content of main or cherry-picking the changes.
- `release` release is to release packages to the public, release channel. This can be used to release from testing or directly from main. Once this has been released it's needed to rebase main and staging to this branch, allowing changes already merged to be synced to the other channels.

When releasing to `release` or `staging` images are not written over if they're already present in the remote channel. I've considered the need for a rewrite of the image a complete edge-case with no clear path forward at the moment. The solution to this is a manual build and manual copy. This can however be changed by adding a flag over the CI that forces the image push, not implemented at the moment. 

TAGS: Tags have been removed from the build process, since I've noted above the need to rewrite history if a botched build is pushed. Releases, however are needed to be made to save the sbom and manifest. 
My proposal is to use TAGS just for releases and tie them up to the release branch instead of the main one.

Note: Not all documentation is up to date, if this draft goes forward I'll look into changing that, not applied at the moment due to uncertainty and changes to be applied if draft gets updated.


docs/build/index.md
```diff
@@ -155,38 +144,28 @@ This ensures all developers and CI get the same default version.
...
+  - Merge any pending documentation PRs and rebuild the docs if needed.
 
-10. **Announce the release:**
-  - Post English announcement on [NethServer Community](https://community.nethserver.org).
-  - Post Italian announcement on [Nethesis Partner Portal](https://partner.nethesis.it).
+6. **Complete the rest of the release tasks:**
+  - Close related issues and milestones.
+  - Archive completed project-board items.
+  - Release NethSecurity Controller if applicable.
+  - Publish the user-facing announcement once the release draft is approved.
```
gsanchietti: Please, re-add both links to forums

docs/design/distfeed.md
```diff
@@ -28,47 +29,54 @@ released and the packages. The repository will receive updates compatible with t
...
+
+It's possible by nature of apk repositories override the variables added by adding the `/etc/apk/repositories.d/98-overrides.list`. This allows to overwrite previously set variables as the user sees fit.
+
+The repository variables are generated in `/etc/apk/repositories.d/99-defaults.list`.
+They include `target_arch`, `package_arch`, `repo_channel`, `version`, and the default `endpoint`.
+If you need to override one of these values, create `/etc/apk/repositories.d/98-overrides.list` and write the variables there.
+
+If you want to change the base URL, add the variable override you need in `98-overrides.list`, `apk` will then handle by itself the replacement.
+
+This changes for image updates, to recieve the correct popup you need to update the `ns-plug.config.repository_url` variable and the `/etc/repo-channel`, this will reflect even into `apk` unless a override has been set. This can be done as the following:
```
gsanchietti: This is not clear, I really do not understand it

docs/design/distfeed.md
```diff
@@ -77,37 +85,36 @@ A machine with a valid subscription receives updates from the subscription chann
 The subscription channel contains stable releases that have undergone additional testing.
 Updates are pushed to the subscription channel after one week from the release date.
 
-If you have a machine with a valid subscription and want to force an update, you can use the following commands:
+If you have a machine with a valid subscription and want to force an update, you can do the following:
 
 ```bash
-cp /etc/apk/repositories.d/customfeeds.list /etc/apk/repositories.d/customfeeds.list.ori
-cat /rom/etc/apk/repositories.d/distfeeds.list | sed 's/dev/stable/g' > /etc/apk/repositories.d/customfeeds.list
+echo 'set repo_channel=staging' > /etc/apk/repositories.d/98-overrides.list
```
gsanchietti: Not sure we should use `staging` and example, I'd use `release`/`stable`.
Use case: an Enterprise user want to install an update directly from Community repo, before it's released to Enterprise.

docs/development_process.md
```diff
@@ -12,45 +12,77 @@ See [NethServer development handbook](https://handbook.nethserver.org/) for a de
...
+- push to `staging` to publish the tested image stream
+- push to `release` to publish the final stable image
 
-See [release new image checklist](../build#release-new-image-checklist) for more information.
+CI will derive the final image name and repository channel automatically. You do not manually copy image artifacts anymore.
 
-### Packages
+### Draft release from a tag
+
+Tags are Git tags in the NethSecurity repository. They are created on the stable release commit, and on GitHub they are visible under the repository tags list and used as the name of the draft release.
```
gsanchietti, line 51: If I want to release a beta release, I should not create any tag, right?
Can we add an example?

gsanchietti (commented):
Minor changes

gsanchietti (approved):

Issue #1547, opened 2026-03-03
ns-ha: possible race condition if switches happen close to one another 


Tbaile:
**Steps to reproduce**

Switch HA role so that the primary firewall (master) resumes operation after a failover. Observe the state of dedalo and dpi services using /etc/init.d/dedalo status and /etc/init.d/dpi status.

**Expected behavior**

After takeover by the primary/master node, dedalo and dpi services should be enabled and remain active as intended, without requiring manual intervention.

**Actual behavior**

Both dedalo and dpi services remain inactive. Manual restart temporarily resolves the issue. Logs from ns-ha show that at failover, restart and stop events overlap: keepalived executes backup_and_stop, then master_and_restart, with commands running too close together and causing the services to be stopped instead of restarted. This is confirmed by observing ns-ha log lines showing restart immediately followed by stop for the same service.

The issue only occurs when the primary takes back over as master. On the slave node, after a switch, services remain up as expected.

Workaround: Manually restart the affected services or schedule a cron job to bring them up if they go down.

**Components**

ns-ha - 0.0.3-r1

**See also**

- Helpdesk ticket: https://helpdesk.nethesis.it/a/tickets/199001
- Internal discussion: https://mattermost.nethesis.it/nethesis/pl/gdi9nek5jpfybgw9h7gjbxyxwh

Tbaile:
We rolled back the change for the moment, to apply the patch:
```
curl --fail https://raw.githubusercontent.com/NethServer/nethsecurity/refs/heads/main/packages/ns-ha/files/00-lock > /etc/hotplug.d/keepalived/00-lock
curl --fail https://raw.githubusercontent.com/NethServer/nethsecurity/refs/heads/main/packages/ns-ha/files/999-unlock > /etc/hotplug.d/keepalived/999-unlock
```

To enable the locking follow the documentation: https://github.com/NethServer/nethsecurity/blob/main/packages/ns-ha/README.md#configuration

ziotibia81:
Issue still preset in version 8.7.2

cotosso:
> Issue still preset in version 8.7.2

We were able to see this behavior only in one specific case, but we were'nt able to find a pattern to reproduce it.
Cause the fix introduces delays in switching we decided to not merge it yet.

If you have any information useful to reproduce the problem please write it here.
Thank you

ziotibia81:
Logs using lock patch:

```
[...]

PR #1548, opened 2026-03-03
fix(ns-ha): added locking mechanism on switch

Tbaile:
Closes: https://github.com/NethServer/nethsecurity/issues/1547

PR #1706, opened 2026-06-03
HA: change for NethSecurtiy 8.8.

gsanchietti:
Main changes:
- re-add lockfile patch from https://github.com/NethServer/nethsecurity/pull/1548
- disable the lock by default, when enabling it, a timeout of 120 seconds is recommended
- move HA monitoring to Victoria Metrics

Fixes: #1547 #1656


packages/telegraf/files/telegraf.conf.d/ha.conf
```diff
@@ -0,0 +1,17 @@
+# High-availability alert state monitoring
+# Exports HA event/state metrics consumed by vmalert.
+
+[[inputs.exec]]
+  name_override = "ha_alert"
+  commands = ["/usr/libexec/telegraf-ha-alert"]
+  interval = "60s"
```
Tbaile, line 7: We could even shrink the interval if we need 🤷

Tbaile (commented):
The existence of the `uci_revert_state` and `uci_set_state` is terrifying.

We could down the like allow telegraf to scrape the service using snmp, until then, fine by me

packages/telegraf/files/telegraf.conf.d/ha.conf
gsanchietti: Sync is done every minute, so if there is a problem the alert should be raised in maximum 2 minutes.
So 60 seconds should be a good default.

PR #1708, opened 2026-06-04
Fix: preserve banip blocklist mtime when dedup content unchanged

gsanchietti:
## Summary

This PR addresses the immediate cause of non-deterministic backup generation: **banip blocklist timestamp issue**.

### 1. banip blocklist timestamp (IMPLEMENTED)

When `ban_deduplicate=1` (the default), banip unconditionally rewrites the blocklist file even when content is unchanged. This updates the mtime and causes backup differences. Fixed by wrapping the write with a `cmp -s` check to only write when content actually changes.

**Files changed:**
- `packages/banip/files/banip-functions.sh` — IPv4 and IPv6 dedup paths now use cmp -s check
- `packages/banip/Makefile` — PKG_RELEASE 3→4

### 2. sysupgrade disable_services and installed_packages timestamps (DEFERRED)

OpenWrt's sysupgrade creates `/etc/uci-defaults/10_disable_services` and `/etc/backup/installed_packages.txt` with the current timestamp. This requires patching upstream OpenWrt base-files which is non-trivial for the v24.10.5 release. This will be addressed as part of upgrading to OpenWrt v25 which includes the upstream fix (OpenWrt PR #16146).

## Related issue

#1146

## How to test

1. Deploy the fixed packages
2. Create two backups on consecutive days: `sysupgrade -k -b /tmp/backup1.tar.gz` and `sysupgrade -k -b /tmp/backup2.tar.gz`
3. Compare: `md5sum /tmp/backup?.tar.gz` — they should now be more consistent (banip blocklist won't cause differences)
4. Verify banip still functions normally when content actually changes
5. Verify sysupgrade backup/restore still works correctly

### DIFF
--- a/packages/banip/Makefile
+++ b/packages/banip/Makefile
@@ -6,7 +6,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=banip
 PKG_VERSION:=1.0.1
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 PKG_LICENSE:=GPL-3.0-or-later
 PKG_MAINTAINER:=Dirk Brenken <dev@brenken.org>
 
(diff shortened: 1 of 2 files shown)


gsanchietti:
Closing in favor of #1709 which targets the correct branch (nethsecurity-8.8 as requested)

PR #1709, opened 2026-06-04
fix(banip): prevent unnecessary blocklist mtime updates

gsanchietti:
## Summary

Fix issue where banip's deduplication process unconditionally rewrites the blocklist file every night, causing system backups to change even when configuration is unchanged.

## Root cause

The banip package (with `ban_deduplicate=1` default) runs nightly and filters the blocklist against a deduplication index. However, it unconditionally writes the result back to `/etc/banip/banip.blocklist`, updating the file's mtime even when the content is identical. Since `sysupgrade -k -b` includes file mtimes in backup tars, this causes non-deterministic backup generation.

## Changes

- Modified `packages/banip/files/banip-functions.sh`:
  - Added `cmp -s` check before writing the deduped blocklist (IPv4 path, line 955-958)
  - Added `cmp -s` check before writing the deduped blocklist (IPv6 path, line 971-975)
  - Now only writes to the blocklist when content actually changes, preserving mtime
- Bumped `packages/banip/Makefile` PKG_RELEASE from 3 to 4

## Testing

- Build passes (GitHub Actions)
- Banip deduplication logic unchanged — only write behavior affected
- File appending paths (auto-add IPs) unaffected

## Related issue

#1146

### DIFF
--- a/packages/banip/Makefile
+++ b/packages/banip/Makefile
@@ -6,7 +6,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=banip
 PKG_VERSION:=1.0.1
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 PKG_LICENSE:=GPL-3.0-or-later
 PKG_MAINTAINER:=Dirk Brenken <dev@brenken.org>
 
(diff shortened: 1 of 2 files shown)


gsanchietti:
While the fix seems to work, it does not really resolve the issue of the backup that always changes MD5 because other files are regenerated on every backup.

PR #1714, opened 2026-06-05
feat(geoblocking): geo ip management in threat shield ip

### LINKED ISSUE #1586: UI support for Geo IP management in Threat Shield IP
**Description**
Add UI support for Geo IP management in the **Threat Shield IP** section.

The new UI must allow administrators to block traffic by geographic area, including both:

* full continent blocking
* granular exceptions within a blocked continent

Example:

* block all Europe
* allow specific countries such as Italy and the UK

The list of supported countries/states must be based on the entries available in the `banip.countries` list:
`https://github.com/openwrt/packages/blob/master/net/banip/files/banip.countries`

Since that list also contains some grouped entries that are not usable as-is, the UI and logic should reorganize the available items according to geographic concepts by comparing the current list with the regional classification provided here:
`https://github.com/lukes/ISO-3166-Countries-with-Regional-Codes/blob/master/all/all.csv`

This feature is needed to make Geo IP filtering easier to configure from the UI and to provide a clearer and more flexible user experience for geographic blocking rules.

It should also include a UI PR for the frontend implementation.

**Proposed solution**
Implement a new **Geo_IP** tab inside **Threat Shield IP**.

The new tab should:

* provide a UI to enable/disable the Geo IP blocking service
* allow selection of whole continents
* allow per-country exceptions and overrides inside selected continents
* support configurations such as “block Europe except Italy and UK”
* use the country list from `banip.countries`
* reorganize countries into meaningful geographic groupings by mapping them against the ISO 3166 regional dataset

Additionally:

* remove the existing **“country block”** toggle currently available among the blacklists
* move that toggle into the new **Geo_IP** tab
* use it as the main global switch for the Geo IP service

The implementation should be delivered with a dedicated PR for the UI side.

**See also**

* `https://github.com/openwrt/packages/blob/master/net/banip/files/banip.countries`
* `https://github.com/lukes/ISO-3166-Countries-with-Regional-Codes/blob/master/all/all.csv`

**Components**
NethSecurity 8.7.2

m-dilorenzi:
This PR contains the implementation of the Geo IP blocking feature in the Threat Shield IP firewall section.
Using this feature allows users to block inbound and outbound connections from specific countries.
Two new APIs, `geoblocking-configuration` and `set-geoblocking-configuration`, have been implemented to allow the UI to retrieve and update the current BanIP configuration for blocked countries.
Each country is categorized using a JSON map that assigns it to the corresponding region (Africa, Oceania, Europe, etc.), which is then shown in the UI.
The `geoblocking-configuration` API returns the state of the Geo IP service (enabled/disabled) and the list of selectable countries, divided by region and indicating whether each one is blocked.
The `set-geoblocking-configuration` API receives the Geo IP service state (enabled/disabled) and the list of blocked countries.
When the service configuration is modified, the backup files `/tmp/banIP-backup/banIP.countryv4.gz` and `/tmp/banIP-backup/banIP.countryv6.gz` are deleted if present. This is necessary because, when the list of blocked countries is changed, banIP uses the IP lists stored in these two backup files, which may contain more or fewer IP addresses than desired.
 
Closes: https://github.com/NethServer/nethsecurity/issues/1586

### DIFF
--- a/packages/ns-api/files/ns.threatshield
+++ b/packages/ns-api/files/ns.threatshield
@@ -240,7 +240,7 @@ continent_table = {
     "SE": "europe",
     "CH": "europe",
     "SY": "asia",
-    "TW": "others",
+    "TW": "asia",
     "TJ": "asia",
     "TZ": "africa",
     "TH": "asia",
--- a/packages/ns-api/files/ns.threatshield
+++ b/packages/ns-api/files/ns.threatshield
@@ -1,7 +1,7 @@
 #!/usr/bin/python3
 
 #
-# Copyright (C) 2024 Nethesis S.r.l.
+# Copyright (C) 2026 Nethesis S.r.l.
 # SPDX-License-Identifier: GPL-2.0-only
 #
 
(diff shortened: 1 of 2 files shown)


packages/ns-api/files/ns.threatshield
```diff
@@ -718,6 +1001,65 @@ def is_ip_blocked():
...
+    }
+
+def set_geoblocking_configuration(e_uci, payload):
+
+    # remove country backup files to avoid stale blocked IPs being restored
+    for backup_file in ['/tmp/banIP-backup/banIP.countryv4.gz', '/tmp/banIP-backup/banIP.countryv6.gz']:
+        try:
+            os.remove(backup_file)
+        except FileNotFoundError:
+            pass
```
Tbaile: Move this into a post-commit, probably cleaner

packages/ns-api/files/ns.threatshield
```diff
@@ -718,6 +1001,65 @@ def is_ip_blocked():
...
+    else:
+        # remove 'country' feed if present
+        if 'country' in feeds:
+            feeds.remove('country')
+    
+    # handle countries if provided
+    if 'countries' in payload:
+        countries_list = payload['countries']
+        if not isinstance(countries_list, list):
+            raise ValidationError('countries', 'invalid', countries_list)
```
Tbaile: Validation should be made before any changes, maybe this is suited to be putted after the `enabled` bool check

packages/ns-api/files/ns.threatshield
```diff
@@ -718,6 +1001,65 @@ def is_ip_blocked():
         'found': False
     }
 
+def geoblocking_configuration(e_uci):
+    try:
+        feeds = list(e_uci.get_all('banip', 'global', 'ban_feed'))
+    except Exception:
+        feeds = []
```
Tbaile: A `get` is probably more accurate here, you can set even defaults, careful that you might still need the `list` cast, since the return for lists is a `tuple`

https://pypi.org/project/pyuci/#:~:text=uci.get(config%2C%20section%2C%20option)

packages/ns-api/files/ns.threatshield
```diff
@@ -718,6 +1001,65 @@ def is_ip_blocked():
...
+    if 'enabled' not in payload:
+        raise ValidationError('enabled', 'required')
+    if not isinstance(payload['enabled'], bool):
+        raise ValidationError('enabled', 'invalid', payload['enabled'])
+    
+    # handle the country feed enable/disable
+    try:
+        feeds = list(e_uci.get_all('banip', 'global', 'ban_feed'))
+    except Exception:
+        feeds = []
```
Tbaile: Same as above

Tbaile (approved):

packages/ns-api/files/ns.threatshield
m-dilorenzi: Moved to `packages/ns-threat_shield/files/adjust-banip.py` for post-commit execution.

packages/ns-api/files/ns.threatshield
```diff
@@ -168,6 +420,34 @@ def get_confidence(f, enterprise=False):
...
+        "americas": [],
+        "asia": [],
+        "europe": [],
+        "oceania": [],
+        "others": []
+    }
+    try:
+        blocked_countries = [c.upper() for c in e_uci.get_all('banip', 'global', 'ban_country')]
+    except Exception:
+        blocked_countries = []
```
Tbaile: Missed an occurrence of the try-catch

Tbaile (approved):
Just a minor leftover, otherwise we good for the backend

PR #1718, opened 2026-06-11
ci: fixed issue with github sha run deduplication

Tbaile:
Just separated the CI in three files, one per channel, build now gets called on workflow_call event.


gsanchietti (commented):
🤷

PR #1734, opened 2026-06-18
feat(dhcp): support multiple dhcp ranges

### LINKED ISSUE #1448: Add support for preserving DHCP ranges configured from terminal
**Brief description**
The UI currently supports only a single DHCP range. If additional DHCP ranges are added manually from the terminal, any subsequent edit in the UI removes all manually configured ranges.

**Why and purpose**
Administrators may need to configure more than one DHCP range, especially in segmented networks or for specific allocation strategies. The inability of the UI to preserve manually added ranges prevents consistent management and forces admins to avoid the UI to prevent accidental configuration loss.

**Steps to reproduce**

1. Configure one DHCP range through the UI.
2. Access the terminal and manually add additional DHCP ranges to the DHCP configuration file.
3. Return to the UI and perform any modification to the DHCP settings.
4. Save changes from the UI.
5. Observe that the manually added DHCP ranges are removed.

**Expected behavior**
The UI should detect and preserve additional DHCP ranges that were added manually, or provide native UI support for configuring multiple DHCP ranges.

**Actual behavior**
Any UI modification overwrites the DHCP configuration file and deletes manually added DHCP ranges.

**Proposed solution**
Improve the UI so that it preserves and respects DHCP ranges added from the terminal, without removing them during save operations. Optionally warn the user when terminal modifications are detected, using a recognizable flag or option to identify those changes.

Version : NethSecurity 8.7.1

m-dilorenzi:
This pull request refactors the DHCP interface handling to support multiple DHCP ranges per network interface in set and get interfaces APIs, ensuring better flexibility and correctness in DHCP configuration.

- updated the data structure for interfaces to use a `ranges` array instead of single `first` and `last` fields, allowing each interface to have multiple DHCP ranges. This affects how interfaces are listed, edited, and retrieved
- added validation to ensure each range is within the interface's subnet, the IPs are non-empty, and the `last` IP is greater than the `first`
- edit logic now deletes existing DHCP configs for the interface and creates new sections for each range
- the `get-interface` API (used to retrieve interface details for editing) now includes `class_start_ip` and `class_end_ip` fields to indicate the valid IP range for UI input box check

Closes: https://github.com/NethServer/nethsecurity/issues/1448

### DIFF
--- a/packages/ns-api/README.md
+++ b/packages/ns-api/README.md
@@ -2732,8 +2732,6 @@ Response example:
 {
   "lan": {
     "device": "br-lan",
-    "start": "",
-    "end": "",
     "active": true,
     "force": true,
     "options": {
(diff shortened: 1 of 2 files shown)


Tbaile (commented):
Just remember to update the docs of ns-api

PR #1738, opened 2026-06-22
fix(dhcp): wrong call of euci for uci prop

### LINKED ISSUE #1694: Preserve dnsmasq DHCP leases on mounted storage

Tbaile:
Addressed issue spotted by @markVnl.

Ref: https://community.nethserver.org/t/switched-to-nethsecurity-on-arm64/26367/34

### DIFF
--- a/packages/ns-api/files/ns.dhcp
+++ b/packages/ns-api/files/ns.dhcp
@@ -222,7 +222,7 @@ def list_active_leases():
         if 'mac' in ldata and 'ip' in ldata:
             static_leases.append(ldata['mac'].lower())
 
-    leasefile = u.get('dhcp', 'ns_dnsmasq', 'dhcp', 'leasefile', default='/tmp/dhcp.leases')
+    leasefile = u.get('dhcp', 'ns_dnsmasq', 'leasefile', default='/tmp/dhcp.leases')
     if not os.path.exists(leasefile):
         return ret
 


gsanchietti (approved):

cotosso:
Just tested on a physical appliance: it works as expected
image

PR #1769, opened 2026-07-06
chore: update banip to v1.8.10

m-dilorenzi:
Update banip package version to v1.8.10

Issue #1897


packages/ns-api-server/files/ns-api-server.initd
```diff
@@ -57,6 +75,11 @@ start_service() {
...
 }
 
+service_triggers() {
+    procd_add_reload_trigger victoria-metrics vmalert
+}
+
 reload_service() {
-   procd_send_signal ns-api-server '*' USR1
+    stop
+    start
```
gsanchietti, line 84: Why this change?

packages/ns-threat_shield/files/banip-extra-defaults
```diff
@@ -0,0 +1,22 @@
...
+	"" | "/var/log/banip.log")
+		uci -q set banip.global.ban_logreadfile="/var/log/messages"
+		;;
+esac
+
+# enable packet counters on the banIP chains and set elements
+if [ -z "$(uci -q get banip.global.ban_nftcount)" ]; then
+	uci -q set banip.global.ban_nftcount="1"
+fi
+
```
gsanchietti, line 19: Call also `ts-ip` just in case Threat Shield subscription is active.
To be moved after commit.

packages/ns-api-server/files/ns-api-server.initd
gsanchietti: Leave as is.

gsanchietti (approved):

PR #1774, opened 2026-07-08
perf(ns-api): speeding up dashboard

### LINKED ISSUE #1819: Dashboard: aggregate API calls into a single request for faster loading on older devices
On older/lower-spec devices the dashboard page felt slow to load: the UI polled roughly 17 separate `ns.dashboard` methods (system info, service statuses, counters, VPN tunnels, etc.) on each refresh, each a separate ubus/rpcd round trip.

**Proposed solution**

Add a `dashboard-v2` aggregate method to `ns.dashboard` that returns all dashboard-page data in a single call: `system`, `services`, `counters`, `vpn`, plus placeholders for `threat_shield`, `mac_binding`, `ips`, `ha`. The internet reachability check is intentionally excluded (it pings remote hosts and can take tens of seconds) — it stays on the dedicated `service-status` method.

Degradation contract: if a subsystem call fails, its key is set to `null` and its name is appended to an `errors` array; the rest of the response is still populated, so one failing subsystem doesn't break the whole dashboard.

**Additional context**

Files/packages to change:

- [`packages/ns-api/files/ns.dashboard`](https://github.com/NethServer/nethsecurity/blob/main/packages/ns-api/files/ns.dashboard) — new aggregate method
- `ns-ui` (nethsecurity-ui) — dashboard page must call the new aggregate endpoint instead of the per-card calls

Tbaile:
Aggregating most of the calls of the dashboard into a single one, allowing the UI to work way smoothly and load up faster on older devices

Precedes: https://github.com/NethServer/nethsecurity-ui/pull/796

Closes #1819


gsanchietti (approved):
While I think it's working, I do not like because there is duplicated code (like __list_events inside ns.snort).

packages/ns-ui/Makefile
```diff
@@ -12,7 +12,7 @@ PKG_RELEASE:=3
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=https://github.com/NethServer/nethsecurity-ui.git
-PKG_SOURCE_VERSION:=$(PKG_VERSION)
+PKG_SOURCE_VERSION:=016db5031f6fd5ef5aa7dbb7aa87a18f515ee1b7
```
gsanchietti: Restore before merge

gsanchietti (approved):
Much better.
Remember to reset PKG_SOURCE_VERSION inside ns-ui before merge

PR #1775, opened 2026-07-08
fix(ns-flashstart): addressing better the bypasses

### LINKED ISSUE #1778: ns-flashstart: bypass rules ignored for network-object entries in Pro+ profiles
**Steps to reproduce**

- Enable FlashStart Pro+ with at least one profile.
- Assign a network object (a CIDR, e.g. `192.168.30.0/24`) — not a single host — to a user in that profile.
- Add that same network (or a range overlapping it) to the FlashStart **Bypass** list.
- Let the `ns-flashstart` daemon run a sync cycle (~25-35s) or restart it, then inspect the resulting ipset, e.g.:
  ```
  nft list set inet fw4 flashstart-ipset-<profile-id>
  ```

**Expected behavior**

Addresses covered by a bypass rule are excluded (fully or partially carved out) from the profile's ipset, so matching traffic is not redirected to FlashStart filtering.

**Actual behavior**

Entries expressed as a network/CIDR (e.g. expanded from a network object assigned to a Pro+ profile user) are never excluded, regardless of whether they overlap a bypass rule. The bypass check (`__is_bypassed_entry`) parsed every entry with `ipaddress.ip_address()`, which raises `ValueError` for any non-host entry and short-circuits to "not bypassed". Whole subnets configured in the bypass list kept being filtered by FlashStart.

**Components**

ns-flashstart `1.0.5`

**See also**

#1579 (earlier bypass fix that did not cover this case)

Tbaile:
Old bypass didn't (or did, and something changed upstream) address completely networks on bypasses.

Closes #1778

### DIFF
--- a/packages/ns-flashstart/files/ns-flashstart
+++ b/packages/ns-flashstart/files/ns-flashstart
@@ -64,18 +64,35 @@ def __load_bypass_rules() -> list:
     return rules
 
 
-def __is_bypassed_entry(entry: str, bypass_rules: list) -> bool:
+def __exclude_bypassed(entry: str, bypass_rules: list) -> list:
+    """Remove any bypassed addresses from entry (a single IP or a CIDR network),
+    carving the bypassed portion out instead of dropping the whole entry.
+    Two CIDR networks are always either disjoint or one fully contains the other,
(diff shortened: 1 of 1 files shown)


gsanchietti (approved):
🤷‍♂️

PR #1779, opened 2026-07-09
fix(threatshield): adding rotation to banIP

Tbaile:
Added banIP to logging too due to banIP using the "suspicious IP" logging to actually count over the offenses.


Tbaile:
Superseded by: https://github.com/NethServer/nethsecurity/pull/1780

PR #1781, opened 2026-07-09
fix(ns-checkmk-utils): sync checkmk-tools v1.7.3 downloads

Coverup20:
## Summary

Sync ns-checkmk-utils with nethesis/checkmk-tools v1.7.3.

This PR supersedes Renovate PR #1773 and includes both:
- the dependency version bump from 0.0.5 to 1.7.3
- the required Makefile inventory/hash updates for the current checkmk-tools layout

## Changes

- Update PKG_VERSION from 0.0.5 to 1.7.3.
- Add check_apk_packages.py to the download, prepare, and install sections.
- Remove check_martian_packets.py.
- Remove check_opkg_packages.py.
- Refresh HASH values against nethesis/checkmk-tools v1.7.3.

## Validation

- Verified all remaining downloads against nethesis/checkmk-tools v1.7.3.
- All HASH values match.
- git diff --check passes.
- Only packages/ns-checkmk-utils/Makefile is changed.

### DIFF
--- a/packages/ns-checkmk-utils/Makefile
+++ b/packages/ns-checkmk-utils/Makefile
@@ -24,7 +24,7 @@ define Download/check_apk_packages
 	URL:=$(NS_CHECKMK_UTILS_BASE_URL)
 	URL_FILE:=check_apk_packages.py
 	FILE:=check_apk_packages-$(PKG_VERSION).py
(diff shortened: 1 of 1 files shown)


gsanchietti:
Replaced by #1784

PR #1791, opened 2026-07-13
fix(nat_helpers): disable modules autoload and update docs

m-dilorenzi:
Disable nathelpers modules autoload after kernel split made on [this PR](https://github.com/openwrt/openwrt/pull/23690).
Each single module has been autoloaded inside the kernel, which can lead to problems since rules have also been created in `nft`.
This PR fixes problems in module retrieval (showed in the UI) and also adds an empty file to override the standard behaviour, preventing modules from being loaded by default.

### DIFF
--- a/docs/design/nat_helpers.md
+++ b/docs/design/nat_helpers.md
@@ -18,9 +18,9 @@ After migration, all NAT helpers are enabled and loaded to preserve NethServer 7
 The helpers are provided as kernel modules by the `kmod-nf-nathelper*` packages.
 List all the available helpers with:
 ```
-for pkg in $(apk list --installed 2>/dev/null | awk '{print $1}' | grep '^kmod-nf-nathelper' | sed 's/-[0-9].*//' | sort -u); do
-    apk info -L "$pkg" 2>/dev/null | grep -e '\.ko$' | sed 's|.*/||;s|\.ko$||'
-done | sort -u
+apk query --installed --format json --fields contents 'kmod-nf-nathelper*' \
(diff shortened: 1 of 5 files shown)


docs/design/nat_helpers.md
```diff
@@ -9,37 +9,36 @@ parent: Design
...
+After migration, all NAT helpers are enabled and loaded to preserve NethServer 7 behavior.
 
-The `kmod-nf-nathelper` package provides the following helpers:
-`apk info -L kmod-nf-nathelper 2>/dev/null | grep -e '\\.ko$' | sed 's|.*/||;s|\\.ko$||'`
+The helpers are provided as kernel modules by the `kmod-nf-nathelper*` packages.
+List all the available helpers with:
 ```
-nf_nat_ftp
-nf_conntrack_ftp
+for pkg in $(apk list --installed 2>/dev/null | awk '{print $1}' | grep '^kmod-nf-nathelper' | sed 's/-[0-9].*//' | sort -u); do
```
Tbaile: Probably this can be simplified using something on the lines of 
```bash
apk query --installed --format json --fields name,contents 'kmod-nf-nathelper*'
```

Tbaile (commented):
Most of the things done using `apk list` can be done through `apk query`, you can then use `--format json` and have a way nicer output to parse instead of going into regex hell, additional documentation can be found here: https://man.archlinux.org/man/apk-query.8.en

Tbaile (approved):

PR #1803, opened 2026-07-16
feat(banner): add name, version and setup hint

filippocarletti:
Add version and suggest to use the "setup" command to the banner shown at login.

### DIFF
--- a/files/etc/banner
+++ b/files/etc/banner
@@ -1,6 +1,0 @@
-
-%D %V
-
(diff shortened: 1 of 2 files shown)


Tbaile:
Produced image wasn't following the desired banner, by digging up a bit I've spotted a nasty issue. If it's cached in the build directory it doesn't get cleaned, the file will stay there. Cached.

PR #1811, opened 2026-07-20
fix(telegraf): improve ping config

gsanchietti:
The native ping method sometime loses ping to the first host inside the configuration.

Improve the config by:

- use exec method so packets for the first host are not lost
- reduce the number of ping
- avoid logging warning messages for the ping plugin

### DIFF
--- a/packages/telegraf/files/telegraf-config
+++ b/packages/telegraf/files/telegraf-config
@@ -25,10 +25,11 @@ PING_TEMPLATE = """# Ping input plugin - monitors ICMP ping to configured hosts
 {% if pings -%}
 [[inputs.ping]]
   urls = [{{ pings|map('tojson')|join(', ') }}]
-  method = "native"
+  method = "exec"
   count = 1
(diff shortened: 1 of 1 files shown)


packages/telegraf/files/telegraf-config
```diff
@@ -25,10 +25,11 @@ PING_TEMPLATE = """# Ping input plugin - monitors ICMP ping to configured hosts
 {% if pings -%}
 [[inputs.ping]]
   urls = [{{ pings|map('tojson')|join(', ') }}]
-  method = "native"
+  method = "exec"
   count = 1
-  ping_interval = 1.0
+  ping_interval = 2.0
   deadline = 10
```
Tbaile: Just this change as doc suggests: https://docs.influxdata.com/telegraf/v1/input-plugins/ping/#configuration
```suggestion
  timeout = "1s"
```

Tbaile (approved):

PR #1814, opened 2026-07-20
fix(ns-checkmk-utils): bump to checkmk-tools v1.7.6, fix production check set

Coverup20:
## Summary
- Bumps `ns-checkmk-utils` from 0.0.5 to 1.7.6, resyncing it with `nethesis/checkmk-tools` (never kept in sync since #1777 was closed unmerged)
- Removes `check_martian_packets`, `check_opkg_packages`, `check_apk_packages`, and `check_uptime` from the install set - none are part of the production check set (see `script-check-nsec8/full/README.md` in checkmk-tools)
- Updates SHA256 hashes for the 10 remaining production checks, verified against both the git tag content and the live `raw.githubusercontent.com` URLs the Makefile fetches
- Adds a postinst that cleans up the 4 excluded scripts (extensionless legacy names and `.py`-suffixed leftovers) left behind by upgrades from older package versions
- Guards the postinst cleanup on `IPKG_INSTROOT`, matching every other package in this repo, so it never runs against the build host filesystem during image builds or offline installs

Replaces #1813, closed to fold in the `IPKG_INSTROOT` guard fix before re-review.

## Test plan
- [ ] Build the package and confirm the 10 production checks install correctly
- [ ] Verify upgrade from a version shipping the 4 excluded scripts removes them via postinst
- [ ] Confirm postinst cleanup is skipped during image build (`IPKG_INSTROOT` set) and runs on live upgrade

### DIFF
--- a/packages/ns-checkmk-utils/Makefile
+++ b/packages/ns-checkmk-utils/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ns-checkmk-utils
 # renovate: datasource=github-tags depName=nethesis/checkmk-tools
-PKG_VERSION:=0.0.5
+PKG_VERSION:=1.7.6
 PKG_RELEASE:=1
 
 PKG_BUILD_DIR:=$(BUILD_DIR)/ns-checkmk-utils-$(PKG_VERSION)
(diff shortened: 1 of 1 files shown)


packages/ns-checkmk-utils/Makefile
```diff
@@ -139,21 +115,39 @@ define Package/ns-checkmk-utils/description
 	tailored for NethSecurity firewall systems.
 endef
 
+# check_apk_packages, check_opkg_packages, check_martian_packets, and
```
gsanchietti, line 118: Please remove this code: the package manager will do it by itself.

gsanchietti (changes requested):

PR #1816, opened 2026-07-20
fix(ns-checkmk-utils): bump to checkmk-tools v1.7.7, fix production check set

Coverup20:
## Summary
- Bumps `ns-checkmk-utils` from 0.0.5 to 1.7.7, resyncing it with `nethesis/checkmk-tools` (never kept in sync since #1777 was closed unmerged)
- Removes `check_martian_packets`, `check_opkg_packages`, `check_apk_packages`, and `check_uptime` from the install set - none are part of the production check set (see `script-check-nsec8/full/README.md` in checkmk-tools)
- Updates SHA256 hashes for the 10 remaining production checks, verified against both the git tag content and the live `raw.githubusercontent.com` URLs the Makefile fetches
- Adds a postinst that cleans up the 4 excluded scripts (extensionless legacy names and `.py`-suffixed leftovers) left behind by upgrades from older package versions
- Guards the postinst cleanup on `IPKG_INSTROOT`, matching every other package in this repo, so it never runs against the build host filesystem during image builds or offline installs
- v1.7.7 (follow-up): checkmk-tools removed the 3 excluded scripts from its `full/` source directory entirely (they were previously kept there only as unused reference) - the 10 packaged production checks are byte-identical to v1.7.6, hashes unchanged

## Test plan
- [x] Build the package and confirm the 10 production checks install correctly
- [x] Verify upgrade from a version shipping the 4 excluded scripts removes them via postinst
- [x] Confirm postinst cleanup is skipped during image build (`IPKG_INSTROOT` set) and runs on live upgrade

### DIFF
--- a/packages/ns-checkmk-utils/Makefile
+++ b/packages/ns-checkmk-utils/Makefile
@@ -115,27 +115,6 @@ define Package/ns-checkmk-utils/description
 	tailored for NethSecurity firewall systems.
 endef
 
-# check_apk_packages, check_opkg_packages, check_martian_packets, and
-# check_uptime are excluded/removed from the production check set (see
-# script-check-nsec8/full/README.md in nethesis/checkmk-tools) but may still
-# be present on disk from a previous package version - clean up both the
-# extensionless legacy names and any .py-suffixed leftovers so an upgrade
-# never leaves stale duplicate CheckMK services behind.
-define Package/ns-checkmk-utils/postinst
-#!/bin/sh
-if [ -z "$${IPKG_INSTROOT}" ]; then
-  rm -f /usr/lib/check_mk_agent/local/check_apk_packages
-  rm -f /usr/lib/check_mk_agent/local/check_apk_packages.py
-  rm -f /usr/lib/check_mk_agent/local/check_opkg_packages
-  rm -f /usr/lib/check_mk_agent/local/check_opkg_packages.py
-  rm -f /usr/lib/check_mk_agent/local/check_martian_packets
-  rm -f /usr/lib/check_mk_agent/local/check_martian_packets.py
-  rm -f /usr/lib/check_mk_agent/local/check_uptime
-  rm -f /usr/lib/check_mk_agent/local/check_uptime.py
-fi
-exit 0
-endef
-
 define Build/Prepare
 	mkdir -p $(PKG_BUILD_DIR)
 	$(CP) $(DL_DIR)/check_dhcp_leases-$(PKG_VERSION).py $(PKG_BUILD_DIR)/check_dhcp_leases.py
--- a/packages/ns-checkmk-utils/Makefile
+++ b/packages/ns-checkmk-utils/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ns-checkmk-utils
 # renovate: datasource=github-tags depName=nethesis/checkmk-tools
-PKG_VERSION:=1.7.6
+PKG_VERSION:=1.7.7
 PKG_RELEASE:=1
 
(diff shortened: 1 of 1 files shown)


packages/ns-checkmk-utils/Makefile
```diff
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ns-checkmk-utils
 # renovate: datasource=github-tags depName=nethesis/checkmk-tools
-PKG_VERSION:=0.0.5
+PKG_VERSION:=1.7.6
```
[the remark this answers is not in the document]
Coverup20: Correct that DEPENDS doesn't declare it explicitly. In practice this isn't an issue for NethSecurity 8.8: python3-nethsec is CONFIG_PACKAGE_python3-nethsec=y in config/python.conf, i.e. it's compiled into every standard NethSecurity 8.8 image regardless of ns-checkmk-utils (which is =m, an optional module). So on any real target build, the library is always present before this package is even installed.
▎
▎ That said, adding +python3-nethsec to DEPENDS would make the package self-describing instead of relying on an implicit base-image guarantee, which matters if it's ever installed on a minimal/custom build. Low priority given the above, but happy to add it if preferred.

packages/ns-checkmk-utils/Makefile
```diff
@@ -139,21 +115,39 @@ define Package/ns-checkmk-utils/description
...
+  rm -f /usr/lib/check_mk_agent/local/check_apk_packages.py
+  rm -f /usr/lib/check_mk_agent/local/check_opkg_packages
+  rm -f /usr/lib/check_mk_agent/local/check_opkg_packages.py
+  rm -f /usr/lib/check_mk_agent/local/check_martian_packets
+  rm -f /usr/lib/check_mk_agent/local/check_martian_packets.py
+  rm -f /usr/lib/check_mk_agent/local/check_uptime
+  rm -f /usr/lib/check_mk_agent/local/check_uptime.py
+fi
+exit 0
+endef
```
gsanchietti: Again, remove

gsanchietti (changes requested):

packages/ns-checkmk-utils/Makefile
Coverup20:
```
# check_apk_packages, check_opkg_packages, check_martian_packets, and
# check_uptime are excluded/removed from the production check set (see
# script-check-nsec8/full/README.md in nethesis/checkmk-tools) but may still
# be present on disk from a previous package version - clean up both the
# extensionless legacy names and any .py-suffixed leftovers so an upgrade
# never leaves stale duplicate CheckMK services behind.

define Package/ns-checkmk-utils/description
      NethSecurity-specific plugins and utilities for the Check_MK agent.
      This package extends the base checkmk-agent package with custom monitoring plugins
      tailored for NethSecurity firewall systems.
endef

# Removes legacy check scripts no longer shipped (see checkmk-tools README)
define Package/ns-checkmk-utils/postinst
      #!/bin/sh
      if [ -z "$${IPKG_INSTROOT}" ]; then
              rm -f /usr/lib/check_mk_agent/local/check_apk_packages
              rm -f /usr/lib/check_mk_agent/local/check_apk_packages.py
              rm -f /usr/lib/check_mk_agent/local/check_opkg_packages
              rm -f /usr/lib/check_mk_agent/local/check_opkg_packages.py
              rm -f /usr/lib/check_mk_agent/local/check_martian_packets
              rm -f /usr/lib/check_mk_agent/local/check_martian_packets.py
              rm -f /usr/lib/check_mk_agent/local/check_uptime
              rm -f /usr/lib/check_mk_agent/local/check_uptime.py
      fi
      exit 0
endef

define Build/Prepare
```
Removed

gsanchietti (approved):

PR #1817, opened 2026-07-20
fix(ns-ha): keep cron jobs disabled on backup across periodic sync

gsanchietti:
## Summary
- HA's periodic (10 min) sync resyncs `/etc/crontabs/root` from the primary to the backup, which silently re-enables the cron lines `update_cron()` had already commented out on the `NOTIFY_BACKUP` transition.
- Re-run `update_cron disable` on `NOTIFY_SYNC` in `600-acme` and `900-ns-plug` so the disable is reapplied every sync cycle instead of drifting back to enabled between failover and the next sync.

## Related issue

#1820

## Test plan
- [x] Verified locally with a stubbed hotplug/crontab harness: `NOTIFY_BACKUP` disables the tracked cron lines, a simulated resync re-enables them, then `NOTIFY_SYNC` re-disables them, leaving unrelated cron lines untouched.
- [x] `shellcheck` on both files (only pre-existing `SC1091`/`SC3014` notices, consistent with the rest of the file).
- [x] Manual verification on a live HA pair (backup node keeps `/etc/crontabs/root` entries commented out across a real sync cycle).

### DIFF
--- a/packages/ns-ha/files/600-acme
+++ b/packages/ns-ha/files/600-acme
@@ -8,7 +8,7 @@ set_service_name acme
 set_restart_if_master
 set_stop_if_backup
 
-if [ "$ACTION" == "NOTIFY_BACKUP" ]; then
+if [ "$ACTION" == "NOTIFY_BACKUP" ] || [ "$ACTION" == "NOTIFY_SYNC" ]; then
     update_cron "disable" "acme"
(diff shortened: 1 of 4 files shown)


Tbaile (approved):
🙈 

Looks fine

PR #1824, opened 2026-07-22
fix(telegraf): skipping down interfaces

Tbaile:
Addressed issue where during HA backup mode the ethtool would keep warning of down interface.


gsanchietti (approved):
Cool

PR #1826, opened 2026-07-24
feat(python3-nethsec): move library and its tests into this repo

gsanchietti:
## What

Moves the `nethsec` Python library (previously the external `NethServer/python3-nethsec` repo) fully into this repository.

- The `nethsec` wheel now builds from local source under `packages/python3-nethsec/src` — no more external fetch.
- Brings the library's tests, test container, and security scans in as GitHub Actions, scoped to this package.
- The library's API reference is now generated inside the existing **Build documentation** workflow and published under the docs site at `/apidocs/python3-nethsec/`.

## Why

Keeps the library, its tests, and its checks in one place so changes here are tested and documented here.

## Checks

- `python3-nethsec-tests` — runs the test suite; builds the test image locally when missing (no registry dependency).
- `python3-nethsec-test-image` — builds and publishes the test image for reuse.
- `python3-nethsec-scans` — Trivy SBOM and report.
- API docs folded into the existing **Build documentation** workflow (the standalone apidocs workflow was removed).

## Verified

- Test image builds; **244 tests pass** (locally and in CI).
- Wheel builds from local source → `nethsec-1.7.1-py3-none-any.whl`.
- pydoctor generates the API HTML cleanly.
- Full OpenWrt image build passed in CI.

## Notes for reviewers

- Version bumps now need both `Makefile` `PKG_VERSION` and `setup.py` `version` (documented in the package README).
- Retiring the old upstream repo is a separate follow-up.


Tbaile (approved):
I'd take the chance to remove the image upload and make the build re-create it every time, we lose a troublesome side that is not fully used at the moment

gsanchietti:
> I'd take the chance to remove the image upload and make the build re-create it every time, we lose a troublesome side that is not fully used at the moment

I agree: removed.

PR #1840, opened 2026-08-03
fix: validate VPN/user names and remove shell interpolation

gsanchietti:
## Summary
- `ns.ovpnrw`: `get_cert_expiration()` now runs `openssl` via argv instead of `shell=True`, and every user-derived OpenVPN PKI file path is resolved through a helper that rejects anything escaping the instance's PKI directory.
- `python3-nethsec`: new `utils.validate_username()` / `utils.validate_uci_name()` reject unsafe characters in user, group and database names before they are written to UCI. Editing/deleting existing records is unaffected, so pre-existing names keep working; only creation of new records is constrained. Bumped to 1.7.2.
- Docs (`ns-objects/README.md`, `ns-api/README.md`) updated to reflect the new `invalid_name` validation error and allowed character sets.

## Test plan
- [x] `packages/python3-nethsec/test.sh` — 259/259 passed, including new validation/regression tests in `test_users.py` and `test_utils.py`.
- [ ] Manual verification on a live device: adding a user/group/database with disallowed characters returns `invalid_name`; existing users with unusual names can still be edited/deleted; OpenVPN cert expiration listing and downloads still work for real users.

### DIFF
--- a/packages/ns-api/Makefile
+++ b/packages/ns-api/Makefile
@@ -6,7 +6,7 @@
 include $(TOPDIR)/rules.mk
  
 PKG_NAME:=ns-api
-PKG_VERSION:=3.7.1
+PKG_VERSION:=3.7.2
 PKG_RELEASE:=1
  
 PKG_BUILD_DIR:=$(BUILD_DIR)/ns-api-$(PKG_VERSION)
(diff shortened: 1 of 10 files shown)


gsanchietti:
Superseded by #1841 (rebased onto integrate-python3-nethsec, pushed from origin instead of fork).

PR #1856, opened 2026-08-03
fix(ns-api): adjusting backup list if none found

### LINKED ISSUE #1855: Backup list API returns inconsistent response shape when no backups exist
**Steps to reproduce**

- Register a firewall with the cloud backup service.
- Ensure the account has zero backups stored remotely.
- Call:
  ```bash
  api-cli ns.backup registered-list-backups
  ```
- Then create at least one backup and call the same method again.

**Expected behavior**

`values` always has the same shape: `{"backups": [...]}`, whether the list is empty or not.

**Actual behavior**

When backups exist:
```json
{"values": {"backups": [{"id": "...", "name": "...", "created": ..., "size": ..., "mimetype": "..."}]}}
```

When no backups exist:
```json
{"values": []}
```

Callers checking for a `backups` key (or iterating `values.backups`) break on the empty case since `values` is a bare array instead of an object.

**Components**

ns-api 3.7.1 (`ns.backup`)
nethsecurity-ui 2.23.2

Tbaile:
Closes: https://github.com/NethServer/nethsecurity/issues/1855

### DIFF
--- a/packages/ns-api/README.md
+++ b/packages/ns-api/README.md
@@ -4274,22 +4274,21 @@ Example response:
 
 ```json
 {
-   "values": [
-      {
-         "file": "882e399c4e6562da3cfa43e886d82454c1b6311392608100801ad0e19307e8b3.bin",
-         "name": "Cool backup"
-      },
-      {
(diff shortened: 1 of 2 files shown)


gsanchietti (commented):
🤷‍♂️

PR #1859, opened 2026-08-04
Update nginx to 1.30.4 and stop it going stale again

gsanchietti:
## Summary

We ship nginx 1.26.3, pinned by the OpenWrt packages feed, missing 12 CVEs. This builds nginx 1.30.4 instead, and wires up the automation that should stop this happening a third time.

Effective nginx version: **1.26.3 → 1.30.4**.

The staleness is not new. The `openwrt-25.12` feed branch is on 1.26.3 and `openwrt-24.10` is on 1.26.1, which never even got the 1.26.3 security fix — OpenWrt release branches get no nginx security backports, so waiting for them is not a plan. We have already tried both alternatives and both froze: the old `packages/nginx` fork was added in Nov 2023 at nginx 1.25.2 and deleted in Mar 2026 still at 1.25.2, and reverting to the feed left us stuck again. Nobody was watching nginx releases, and a stale version never fails a build, so it stays invisible.

So the fix is not another mechanism change:

- nginx 1.30.4 via a version-only feed patch, kept at a filename without the version in it so Renovate can edit it in place
- a Renovate custom manager that reads nginx's own releases (`github.com/nginx/nginx`, `release-*` tags), bounded to the 1.30.x stable branch, so every future nginx release arrives as a PR
- the policy written down in `docs/build/index.md`: which packages we track ahead of OpenWrt, why stable-branch-only, and that deleting the patch to unbreak a build after an `OWRT_VERSION` bump is a silent downgrade

Renovate cannot compute `PKG_HASH`, so its PRs will fail the build until someone refreshes the hash. That failure is the notification, and the command to run is in the patch header.

## How to test

Check the build log: nginx 1.30.4 downloads, all ten feed patches apply, and `nginx-ssl`, `nginx-mod-luci`, `nginx-mod-ubus`, `nginx-mod-naxsi` and `nginx-mod-headers-more` compile.

On a device flashed with the PR image:

- `nginx -v` reports 1.30.4 and `nginx -t` is clean
- the web UI loads over HTTPS and `api-cli ns.dashboard system-info` works
- a reverse proxy route still reaches its backend
- upgrade an existing 1.26.3 device, not just a fresh flash

This jumps two nginx stable branches, so watch for behaviour changes: backend connections are now keepalive and HTTP/1.1 by default (1.29.7), `Host` and request-line validation follows RFC 3986 (1.29.4), and TLS 1.0/1.1 are off by default (1.27.3). Future 1.30.x bumps are security-only and should not drift.

CVEs closed: CVE-2026-42926, -42945, -42946, -42934, -40460, -40701, -9256, -42055, -48142, -42533, -60005, -56434.


gsanchietti:
Dropped for now

PR #1865, opened 2026-08-06
fix(ns-api): write /tmp/TZ without a shell

gsanchietti:
The timezone post-commit hook passed the UCI timezone value through "sh -c echo '<value>' > /tmp/TZ", so a value containing a single quote broke out of the quoting and the rest was executed as a command. Write the file directly with Python instead: no shell is involved, and the resulting content is unchanged.

Assisted-by: Claude Code:claude-opus-5[1m]


m-dilorenzi:
LGTM

image
image

m-dilorenzi (approved):

PR #1868, opened 2026-08-06
fix: few repo adjustments

### LINKED ISSUE #1838: Installing upstream OpenWrt packages fails due to malformed version string
**Steps to reproduce**

- On a NethSecurity 8.8.0 device, enable upstream OpenWrt package repositories as documented (`/etc/apk/repositories.d/customfeeds.list` using `${openwrt_version}`)
- Run `apk update` / attempt to install any package from that feed

**Expected behavior**

The package downloads and installs successfully using the OpenWrt release path (e.g. `25.12.5`).

**Actual behavior**

The fetch fails with `wget: exited with error 8` (not found). The `openwrt_version` apk variable is published with a `v` prefix (e.g. `v25.12.5`), but upstream OpenWrt release paths never use that prefix, so the constructed URL is invalid.

**Components**

ns-plug 1.3.1 (`distfeed-setup` script)
https://community.nethserver.org/t/upstream-openwrt-packages-installation-not-working/28220

### LINKED ISSUE #1867: Automatic package updates bypass custom feeds exclusion
**Steps to reproduce**

- Add a custom feed via `customfeeds.list` (see `docs/design/distfeed.md`).
- Wait for the nightly package-update cron to run, or trigger a package update check/install from the UI.

**Expected behavior**

Automatic update paths (nightly cron and UI update flow) should ignore `customfeeds.list` and only consider packages from the official NethSecurity distfeed, since custom-feed packages are not rebuilt or QA'd by NethSecurity.

**Actual behavior**

Nightly updates and the UI's update check/install flow call `apk` directly, so custom feeds are considered like any other repo. This can cause a package to be upgraded from an upstream/custom feed by mistake, bypassing NethSecurity's QA channel. Manual `apk` invocations are unaffected either way.

**Components**

ns-plug 1.3.1
ns-api 3.7.1

Tbaile:
- **fix(ns-plug): trimming initial v from the openwrt version**: closes https://github.com/NethServer/nethsecurity/issues/1838
- **fix: not using custom feeds when automatic updating**: closes https://github.com/NethServer/nethsecurity/issues/1867
- **build: compiling nano and htop**: requested by community: https://community.nethserver.org/t/upstream-openwrt-packages-installation-not-working/28220/5


config/utils.conf
```diff
@@ -0,0 +1,2 @@
+CONFIG_PACKAGE_nano-full=m
+
```
gsanchietti: Remove i possible

packages/ns-plug/files/apk-official
```diff
@@ -0,0 +1,26 @@
...
+# excluded, so automatic update paths never install a package that bypassed
+# the NethSecurity distfeed/QA channel. Serialized via flock so overlapping
+# invocations (cron, UI, manual admin apk use) don't race on the file.
+
+LOCK=/var/run/apk-official.lock
+CUSTOMFEEDS=/etc/apk/repositories.d/customfeeds.list
+DISABLED="${CUSTOMFEEDS}.disabled"
+
+exec 9>"$LOCK"
+flock 9
```
gsanchietti, line 19: Why flock is needed?
apk should already handle the lock

gsanchietti: Confirmed: apk handles the lock.
When another instance is running, this is the error that apk returns:
```
ERROR: Unable to lock database: Resource temporarily unavailable
ERROR: Failed to open apk database: Resource temporarily unavailable
```

gsanchietti (changes requested):
Remove the flock call

packages/ns-plug/files/apk-official
Tbaile: I'm not concerned about `apk`, more about the file move. We could lose the custom feeds if concurrent requests are being made, if it's acceptable, I'll remove it no problem

gsanchietti (approved):

Issue #1869, opened 2026-08-06
Firewall changes are silently ignored on large configurations


gsanchietti:
## Component and version

NethSecurity 8.8.0 — `firewall4` (`firewall4-2025.03.17~b6e51575-r2`)

## Actual behavior

On firewalls with many port forwards and many zones, applying the firewall configuration fails halfway through building the rules. The new rules are then **never loaded**: rules are applied all-or-nothing, so the incomplete set is discarded and the firewall keeps enforcing its previous rules. The command still reports success.

The practical consequence: an administrator can close a port, remove an access rule, or add a port forward, see no error at all, and reasonably believe the change is active — when it is not. The firewall keeps applying the old rules until someone notices.

`fw4 reload` prints:

```
Runtime error: Unable to open source file /usr/share/firewall4/templates/redirect.uc: No file descriptors available
In main(), file /usr/share/firewall4/templates/ruleset.uc, line 338, byte 53:
  called from function include ([C])
  called from function render_ruleset (/usr/share/firewall4/main.uc:58:72)

/dev/stdin:1698:3-3: Error: syntax error, unexpected end of file
```

Note that `fw4 reload` still exits with status **0**, so scripts and automation cannot detect the failure.

Seen on a production firewall with 315 configuration sections (66 port forwards, 23 zones).

## Expected behavior

Either the configuration is applied successfully, or the command fails visibly with a non-zero exit status.

## Cause

While building the rules, one file is opened for each rule template used, and none are closed until the whole set is finished. Every port forward also multiplies this, because each "reflection" zone listed on a port forward is rendered separately.

Large configurations therefore exceed the limit of 1024 open files per process. Measured on the test configuration below: **1060 files needed against 1024 allowed** — only ~36 over. This means affected systems sit just barely over the threshold, and adding a single port forward can be enough to trigger it. It also makes the problem easy to miss in testing: one fewer port forward, or a WAN interface without an address, keeps it just under the limit and everything appears fine.

## Steps to reproduce

Requires a test machine with a spare, unused network card, and a WAN interface that has an IP address.

1. Save the script below as `/tmp/gen.sh` and run it (it backs up the current configuration to `/root/firewall.bak` first):

   ```sh
   sh /tmp/gen.sh
   ```

2. Apply the configuration:

   ```sh
   fw4 reload
   ```

3. The error above is printed. Confirm the new rules were *not* applied — the count is unchanged from before the reload:

   ```sh
[...]

PR #1870, opened 2026-08-06
Raise fw4 open-file limit

gsanchietti:
## Summary

On firewalls with many port forwards and zones, applying the firewall
configuration failed part way through. The incomplete rules were then
discarded, so the firewall kept its previous rules while the command still
reported success — changes appeared to work but did not take effect.

Building the rules holds one file open per rule template until the whole set
is finished, which exceeds the limit of 1024 open files. The patch raises that
limit in `/sbin/fw4`, the single place every reload goes through (boot,
hotplug, API, manual runs).

`PKG_RELEASE` is bumped too, so the rebuilt package counts as an upgrade and
can be shipped to existing installations without releasing a new image.

Related issue: https://github.com/NethServer/nethsecurity/issues/1869

## How to test

Issue #1869 has a script that builds a firewall configuration large enough to
trigger the problem. On a test machine, run it and then `fw4 reload`:

- current release: fails with `No file descriptors available`, and the new
  rules are not applied (`nft list table inet fw4 | grep -c "dnat ip to"` is
  unchanged)
- with this patch: completes, and all 264 rules are loaded

Note `fw4 reload` exits 0 in both cases, so check the printed output rather
than the exit status.

To test on a real machine:
```
cat > /etc/apk/repositories.d/98-overrides.list << "EOF"
set -default repo_channel=pr1870
set -default version=8.8.0-pr1870.342.20260806142738.91c2eae
EOF
apk update
apk upgrade firewall4
fw4 reload <- error must be gone
```

## Still to confirm

Whether the `distfeed.nethesis.it` subscription mirror carries the `base`
feed. If it only mirrors `nethsecurity`, subscription units would need a new
image and this should get the `needs image` label.

### DIFF
--- a/patches/package/network/config/firewall4/20-raise-nofile-ulimit.patch
+++ b/patches/package/network/config/firewall4/20-raise-nofile-ulimit.patch
@@ -1,0 +1,36 @@
+# Raise fw4's open-file limit so large rulesets can be applied.
+# PKG_RELEASE is bumped so this ships as a package update, without a new image.
+# https://github.com/NethServer/nethsecurity/issues/1869
+--- firewall4.ori/Makefile
++++ firewall4/Makefile
+@@ -5,7 +5,7 @@
+ include $(TOPDIR)/rules.mk
+ 
+ PKG_NAME:=firewall4
+-PKG_RELEASE:=2
++PKG_RELEASE:=3
+ 
+ PKG_SOURCE_PROTO:=git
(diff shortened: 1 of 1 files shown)


Tbaile (approved):
Really disliking this, but I feel there's no solution

gsanchietti:
Upstream is doing the fix, let's wait for it.

PR #1875, opened 2026-08-07
feat(adblock): implement device-based enforcement and migration scripts

### LINKED ISSUE #1871: LAN DNS traffic is no longer redirected to dnsmasq
**Steps to reproduce**

- Enable Threat Shield DNS from the UI (`Security > Threat Shield DNS`) selecting the `lan` zone
- Check the resulting configuration: `uci show adblock.global.adb_nftdevforce` → `'lan'`
- Resolve the actual device behind that zone: the zone `lan` maps to network `green`, which maps to device `eth0`
- Inspect the generated ruleset: `nft list chain inet adblock pre-routing`
- From a LAN client, send a DNS query to an external resolver, bypassing the firewall's DNS: `dig @8.8.8.8 -x <client_ip> +short`

**Expected behavior**

DNS traffic from the selected zones is redirected to the local dnsmasq instance on port 53, and DoT on port 853 is rejected, so that clients configured with a custom DNS server are still filtered by Threat Shield DNS.

**Actual behavior**

The enforcement rules are generated using the **firewall zone name** as `iifname`. Since `iifname` matches on the actual network device name, the rules never match any packet and the counters stay at zero:

```
root@NethSec:~# nft list chain inet adblock pre-routing
table inet adblock {
	chain pre-routing {
		type nat hook prerouting priority mangle; policy accept;
		iifname "lan" meta nfproto { ipv4, ipv6 } meta l4proto { tcp, udp } th dport 53 counter packets 0 bytes 0 redirect to :53
		iifname "lan" meta nfproto { ipv4, ipv6 } meta l4proto { tcp, udp } th dport 853 counter packets 0 bytes 0 goto _reject
	}
}
```

The value should be a network device (`eth0` in this setup), not a zone name:

```
adblock.global.adb_nftdevforce='lan'    # actual
adblock.global.adb_nftdevforce='eth0'   # expected
```

**Workaround**

Add the enforcement rules manually, using the real device name instead of the zone name. Retrieve the device of each zone you want to enforce:

```sh
. /lib/functions/network.sh
network_get_device dev green && echo "$dev"
```

Then add the rules to the existing chain (replace `eth0` with the device from the previous command):

```sh
nft add rule inet adblock pre-routing iifname "eth0" \
    meta nfproto { ipv4, ipv6 } meta l4proto { udp, tcp } th dport 53 counter redirect to :53
nft add rule inet adblock pre-routing iifname "eth0" \
    meta nfproto { ipv4, ipv6 } meta l4proto { udp, tcp } th dport 853 counter goto _reject
```

Verify with `nft list chain inet adblock pre-routing` that the counters start increasing.

Caveat: these rules live in memory only. They are lost on reboot and every time adblock is reloaded or restarted — including the scheduled cron reload — so they need to be re-applied, e.g. from a hotplug/startup script.

m-dilorenzi:
This PR fixes the issue where DNS queries are no longer redirected to dnsmasq after the adblock upgrade to 4.5.5.

Up to adblock 4.1.5 the local DNS enforcement was configured with firewall zone names, and adblock turned them into UCI redirect sections letting fw4 resolve each zone into its devices. Since 4.5.5 adblock writes nftables rules directly and expects the `adb_nftdevforce` option to contain device names, because its values are used as-is in the `iifname` match. Zone names kept being stored there, so the resulting rules never matched any packet and DNS traffic was not intercepted.

The fix makes `adb_nftdevforce` a derived value: the zones selected in the UI are stored in a dedicated `ns_tsdns_zones` option and expanded into the devices of those zones. The UI is not affected, it keeps working with zones.

### Changes

- `packages/ns-api/files/ns.threatshield`: `dns_edit_settings` now saves the selected zones in `ns_tsdns_zones` and writes the devices they contain into `adb_nftdevforce`; `dns_list_settings` reads the zones back from the new option, with a fallback that keeps the settings page working until the migration has run.
- `packages/ns-threat_shield/files/configure-adblock-devices.py` (new): pre-commit hook that recomputes `adb_nftdevforce` whenever `adblock`, `firewall` or `network` change, so that adding an interface to an enforced zone does not leave its DNS traffic unfiltered.
- `packages/ns-threat_shield/files/adblock-devices-migrate.py` (new): rebuilds the configuration on systems already upgraded, which would otherwise stay broken until someone re-saves the Threat Shield DNS settings, and restarts adblock.
- `packages/ns-threat_shield/files/36_ns-threat_shield` (new): uci-defaults running the migration above, following the same pattern used for the wireguard migration.
- `packages/ns-threat_shield/Makefile`: installs the new hook, migration and uci-defaults, and raises the `ns-api` requirement so the migration cannot be undone by an older API still writing zone names.

Closes: https://github.com/NethServer/nethsecurity/issues/1871

### DIFF
--- a/packages/ns-api/files/ns.threatshield
+++ b/packages/ns-api/files/ns.threatshield
@@ -736,24 +736,12 @@ def dns_list_zones(e_uci):
         zones.append(zone['name'])
     return { 'data': zones }
 
-def dns_get_enforced_zones(e_uci):
-    # the user selection is stored in ns_tsdns_zones: adb_nftdevforce holds the
-    # devices derived from it and can't be mapped back to zones without ambiguity
-    zones = list(e_uci.get('adblock', 'global', 'ns_tsdns_zones', list=True, default=[]))
-    if zones:
-        return zones
-    # before the switch to device-based enforcement the selection was stored inside
-    # adb_nftdevforce: keep reading it until the migration has run
-    return list(e_uci.get('adblock', 'global', 'adb_nftdevforce', list=True, default=['lan']))
-
 def dns_zones_to_devices(e_uci, zones):
     # adb_nftdevforce is rendered as an nft iifname match, so it needs the devices
     # of the selected zones and not the zone names
-    devices = []
+    devices = set()
     for zone in zones:
-        for device in utils.get_all_devices_by_zone(e_uci, zone, exclude_aliases=True):
-            if device not in devices:
-                devices.append(device)
+        devices.update(utils.get_all_devices_by_zone(e_uci, zone, exclude_aliases=True))
     return sorted(devices)
 
 def dns_set_enforced_devices(e_uci, zones):
@@ -766,7 +754,11 @@ def dns_set_enforced_devices(e_uci, zones):
 
 def dns_list_settings(e_uci):
     ts_enabled = e_uci.get('adblock', 'global', 'ts_enabled', default='0')
-    zones = dns_get_enforced_zones(e_uci)
+    # the user selection is stored in ns_tsdns_zones: adb_nftdevforce holds the devices
+    # derived from it and can't be mapped back to zones without ambiguity. Installations
+    # predating the device-based enforcement are converted by 96_ns-threat_shield, so the
+    # option is always set at this point
+    zones = list(e_uci.get('adblock', 'global', 'ns_tsdns_zones', list=True, default=['lan']))
     try:
         ports = list(e_uci.get_all('adblock', 'global', 'adb_nftportforce'))
     except:
--- a/packages/ns-api/files/ns.threatshield
+++ b/packages/ns-api/files/ns.threatshield
@@ -736,12 +736,37 @@ def dns_list_zones(e_uci):
         zones.append(zone['name'])
     return { 'data': zones }
 
+def dns_get_enforced_zones(e_uci):
+    # the user selection is stored in ns_tsdns_zones: adb_nftdevforce holds the
+    # devices derived from it and can't be mapped back to zones without ambiguity
+    zones = list(e_uci.get('adblock', 'global', 'ns_tsdns_zones', list=True, default=[]))
+    if zones:
+        return zones
+    # before the switch to device-based enforcement the selection was stored inside
+    # adb_nftdevforce: keep reading it until the migration has run
+    return list(e_uci.get('adblock', 'global', 'adb_nftdevforce', list=True, default=['lan']))
+
+def dns_zones_to_devices(e_uci, zones):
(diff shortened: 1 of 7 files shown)


gsanchietti (approved):

packages/ns-api/files/ns.threatshield
```diff
@@ -736,12 +736,37 @@ def dns_list_zones(e_uci):
...
+
+def dns_zones_to_devices(e_uci, zones):
+    # adb_nftdevforce is rendered as an nft iifname match, so it needs the devices
+    # of the selected zones and not the zone names
+    devices = []
+    for zone in zones:
+        for device in utils.get_all_devices_by_zone(e_uci, zone, exclude_aliases=True):
+            if device not in devices:
+                devices.append(device)
+    return sorted(devices)
```
Tbaile, line 757: Next time you can use a `set`, so you don't need to check for presence in the list, for the number of occurrences, do not bother changing it

packages/ns-api/files/ns.threatshield
```diff
@@ -736,12 +736,37 @@ def dns_list_zones(e_uci):
...
 
+def dns_get_enforced_zones(e_uci):
+    # the user selection is stored in ns_tsdns_zones: adb_nftdevforce holds the
+    # devices derived from it and can't be mapped back to zones without ambiguity
+    zones = list(e_uci.get('adblock', 'global', 'ns_tsdns_zones', list=True, default=[]))
+    if zones:
+        return zones
+    # before the switch to device-based enforcement the selection was stored inside
+    # adb_nftdevforce: keep reading it until the migration has run
+    return list(e_uci.get('adblock', 'global', 'adb_nftdevforce', list=True, default=['lan']))
```
Tbaile, line 747: After the uci default, this is not needed, right?
We can safely remove it and put this logic underneath

packages/ns-threat_shield/Makefile
```diff
@@ -56,9 +56,13 @@ define Package/ns-threat_shield/install
 	$(INSTALL_DATA) ./files/banip.nethesis.feeds $(1)/etc/banip
 	$(INSTALL_BIN) ./files/adjust-banip.py $(1)/usr/libexec/ns-api/post-commit/
 	$(INSTALL_BIN) ./files/configure-banip-wans.py $(1)/usr/libexec/ns-api/pre-commit/
+	$(INSTALL_BIN) ./files/configure-adblock-devices.py $(1)/usr/libexec/ns-api/pre-commit/
+	$(INSTALL_DIR) $(1)/usr/libexec
```
Tbaile, line 60: Move it at the top where all install dir resides, this missing could've cause issues.

Tbaile (changes requested):

Tbaile (approved):

PR #1882, opened 2026-09-01
chore(netifyd): updated to 5.2.9

### LINKED ISSUE #1908: netifyd chores: update to 5.2.9 and add ARM64 support
Bump the netifyd DPI agent to the latest upstream release, cleanup unused plugins, deploy (for the moment the build is private) ARM64 bins.

- Remove the device-discovery processor plugin and SQLite sink plugin, they were never used, was only built for testing purposes
- Remove socket sink plugins since it's not needed anymore, moved to internal HTTP transmissions
- Disable the HTTP sink plugin by default, tools that use it drop in their configuration
- Tune netlink buffer settings to upstream's new defaults (buffer size, max buffer size, socket buffer size, conntrack idle timeout)
- Add ARM64 (aarch64) build/package support for netifyd

Tbaile:
Removed old sync script and added a new one.
Updated to netifyd 5.2.9.
Removed unused plugins.
Added support for ARM builds.


tools/netifyd-update/netifyd-update.py
```diff
@@ -0,0 +1,248 @@
...
+from pathlib import Path
+
+SCRIPT_DIR = Path(__file__).resolve().parent
+DEFAULT_WORK_DIR = SCRIPT_DIR / "work"
+DEFAULT_DIST_DIR = SCRIPT_DIR / "dist"
+NETIFYD_PKG_DIR = SCRIPT_DIR.parent.parent / "packages" / "netifyd"
+DIST_MK = NETIFYD_PKG_DIR / "netify-dist.mk"
+
+SOURCES: dict[str, str] = {
+    "x86": "https://download.netify.ai/5/openwrt/25.12/x86/index.json",
```
gsanchietti, line 23: maybe 25.12 can be calculated from an env var from the build system?

Tbaile: I tried, but that field will likely be made change various times, this is a sync script, manual work will be needed to check that you're downloading the correct version. We could extract it from the `build.conf` and `build.conf.override` but you still need to manually upload the files to the S3 before continuing, so very little improvement could come from that. 

(for example, while building i've added the arm builds that are available only in our integrator branch, had to remove them from the file since they're not publicly available yet)

gsanchietti (approved):
🤷‍♂️ 

does dev discovery plugin get preserved? I see the config is gone

Tbaile:
Nope, gets :axe:'d with socket and sqlite too.

We don't use them, it's better for the moment to not have them since it's more files to maintain.

gsanchietti (commented):
🤷‍♂️

Tbaile:
All ready, waiting for planned package releases before merge.

PR #1883, opened 2026-09-02
feat(inventory): add geoblocking status

gsanchietti:
Track geoblocking feature usage.
Values meaning:
- 0 no geoblocking enabled
- \>=1 geoblocking enabled, number of blocked countries

### DIFF
--- a/packages/python3-nethsec/src/nethsec/inventory/__init__.py
+++ b/packages/python3-nethsec/src/nethsec/inventory/__init__.py
@@ -394,14 +394,16 @@ def fact_threat_shield(uci: EUci):
     ret = { 'enabled': False, 'community': 0, 'enterprise': 0, 'geoblocking': 0 }
     ret['enabled'] = uci.get('banip', 'global', 'ban_enabled', default='0') == '1'
     try:
-        for feed in uci.get_all("banip", "global", "ban_feed"):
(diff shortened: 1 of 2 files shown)


packages/python3-nethsec/src/nethsec/inventory/__init__.py
```diff
@@ -401,6 +401,7 @@ def fact_threat_shield(uci: EUci):
                 ret['community'] += 1
     except:
         pass
+    ret['geoblocking'] = len(uci.get('banip', 'global', 'ban_country', list=True, default=[]))
```
m-dilorenzi: Checking the `banip.global.ban_country` property is not enough. The correct way to check whether GeoIP blocking is active is to check `banip.global.ban_feed`, which must contain the `country` feed. The `banip.global.ban_country` property can also be set when GeoIP blocking is not active; this behavior was implemented to let users keep their previous country selection when re-enabling GeoIP blocking.

m-dilorenzi (changes requested):

packages/python3-nethsec/src/nethsec/inventory/__init__.py
gsanchietti: Fixed

m-dilorenzi (approved):

PR #1889, opened 2026-09-08
feat(dpi): refactor DPI section

m-dilorenzi:
This PR contains a rework on the DPI rule management around application groups and an ordered, first-match-wins rule list, replacing the old flat schema where every rule was a device plus a list of applications or protocols, all evaluated in parallel.

### Main changes

- **Application groups** as a new reusable entity, referenced by rules instead of inlining applications and protocols in each one.
- **Rules** now carry a name, an action, an optional source and a priority, and halt on match, so their order decides the outcome. Two actions: `block` and the new `allow`.
- **API**: rule CRUD is id-based and gains rename, enable, disable and reorder, plus the application group CRUD. Catalog lookups are split between what the engine has loaded and the downloaded catalogs. Removed `list-popular`, `list-devices` and the global exemption CRUD, replaced by an Allow rule at the top of the list.
- **Migration**: a uci-defaults script converts `/etc/config/dpi` on the first boot after the upgrade. Legacy rules are kept with their behaviour frozen verbatim, global exemptions become Allow rules, unsupported rules are dropped with a log line naming them.
- **Generator**: `dpi-config` orders the actions by priority, refuses to emit a rule that would produce an invalid criteria, and still handles the old schema for a box whose migration has not run yet.


packages/python3-nethsec/src/nethsec/dpi/__init__.py
```diff
@@ -221,276 +229,1000 @@ def list_applications(search: str = None, limit: int = None, page: int = 1) -> d
...
+def __apply(e_uci: EUci):
+    """
+    Commit the dpi config and reload the dpi service.
+
+    Used by the schema migration only: it runs at boot with nobody around to confirm the pending uci
+    changes, so it must take effect on its own. Rule and appgroup CRUD go through the normal pending
+    uci changes instead, applied later by ns.commit like every other section.
+    """
+    e_uci.commit('dpi')
+    subprocess.run(["/etc/init.d/dpi", "reload"], check=True)
```
Tbaile: I used only by migration, should stay inside migration function

packages/python3-nethsec/src/nethsec/dpi/__init__.py
```diff
@@ -221,276 +229,1000 @@ def list_applications(search: str = None, limit: int = None, page: int = 1) -> d
...
+def __toggle_engine(e_uci: EUci):
+    count_enabled = 0
+    for section in e_uci.get_all('dpi'):
+        if e_uci.get('dpi', section, default="") == "rule" and e_uci.get('dpi', section, 'enabled', default="0") == "1":
+            count_enabled = count_enabled + 1
+
+    if count_enabled > 0:
+        e_uci.set('dpi', 'config', 'enabled', '1')
+    else:
+        e_uci.set('dpi', 'config', 'enabled', '0')
```
Tbaile: This was cause netifyd was active only when used, now it's active all the time, we can remove this and adapt the scripts accordingly.

packages/python3-nethsec/src/nethsec/dpi/__init__.py
```diff
@@ -221,276 +229,1000 @@ def list_applications(search: str = None, limit: int = None, page: int = 1) -> d
...
+        elif kind == 'proto':
+            vocabulary |= {value.lower() for value in load_protocols().values()}
         else:
-            result.append({
-                'name': popular_filter,
-                'missing': True
-            })
+            category_type = 'application' if kind == 'app_category' else 'protocol'
+            vocabulary |= {value.lower() for value in load_categories().get(category_type, {}).values()}
+    except Exception:
```
Tbaile: set is, by definition, a unique list of elements, the bitwise operator on it kinda doesn't make any sense, what does it fix?

packages/python3-nethsec/src/nethsec/dpi/__init__.py
```diff
@@ -221,276 +229,1000 @@ def list_applications(search: str = None, limit: int = None, page: int = 1) -> d
...
+        return None
+    if objects.is_object_id(criteria):
+        return objects.get_object_ips(e_uci, criteria) or None
+    try:
+        return __validate_source([criteria])
+    except ValidationError:
+        return None
+
+
+def migrate_schema(e_uci: EUci) -> bool:
```
Tbaile: why the bool return?

PR #1890, opened 2026-09-10
feat: added controller attribution

Tbaile:


packages/ns-api-server/files/src/middleware/middleware.go
```diff
@@ -30,13 +30,57 @@ import (
...
+
+// checkOnBehalfOf returns onBehalfOf when username is the controller machine account and the value is valid, or "" otherwise.
+func checkOnBehalfOf(onBehalfOf string, username string) string {
+	onBehalfOf = strings.TrimSpace(onBehalfOf)
+	if onBehalfOf == "" || len(onBehalfOf) > onBehalfOfMaxLen {
+		return ""
+	}
+
+	// control characters would let a crafted name forge extra log lines
+	if strings.ContainsFunc(onBehalfOf, func(r rune) bool { return r < 0x20 || r == 0x7f }) {
```
gsanchietti, line 55: I'd move to a regexp validation since onBehakfOf is always a unix-like user: something like `regexp.MustCompile(`^[A-Za-z0-9._@-]+$`)`

packages/ns-api-server/files/src/middleware/middleware.go
```diff
@@ -89,22 +136,24 @@ func InitJWT() *jwt.GinJWTMiddleware {
...
+				// create claims map
+				claims := jwt.MapClaims{
 					identityKey: user.Username,
 					"role":      "",
 					"actions":   []string{},
 					"2fa":       status == "1",
 				}
+
+				if user.SudoRequested {
+					claims["sudo"] = time.Now().Unix()
```
gsanchietti, line 148: Do you think it could be useful to log the sudo request?

Issue #1899, opened 2026-09-15
OpenVPN P2P tunnel fails to start


francio87:
OpenVPN net-to-net tunnels configured in P2P mode fail to start.

**Steps to reproduce**

- Configure an OpenVPN net-to-net tunnel in P2P mode on a NethSecurity firewall (fw-server).
- Export the generated JSON configuration.
- Import the JSON configuration on a second NethSecurity firewall (fw-client).
- Enable/start the tunnel on both firewalls.
- Check the OpenVPN service status and logs.

**Expected behavior**

The OpenVPN net-to-net tunnel configured in P2P mode should start successfully and establish the connection between the two firewalls.

**Actual behavior**

The tunnel does not start on either side

Tunnel Server logs :

```
Sep 15 11:11:34 fw-server openvpn(ns_testp2p)[21639]: DEPRECATED OPTION: --persist-key option ignored. Keys are now always persisted across restarts. 
Sep 15 11:11:34 fw-server openvpn(ns_testp2p)[21639]: DEPRECATED OPTION: The option --secret is deprecated.
Sep 15 11:11:34 fw-server openvpn(ns_testp2p)[21639]: Note: Using --push without --mode server is an unsupported configuration. Negotiation of OpenVPN features is expected to fail.
Sep 15 11:11:34 fw-server openvpn(ns_testp2p)[21639]: Options error: DEPRECATION: No tls-client or tls-server option in configuration detected. OpenVPN 2.8 will remove the functionality to run a VPN without TLS. See the examples section in the manual page for examples of a similar quick setup with peer-fingerprint. OpenVPN 2.7 allows using this configuration when using --allow-deprecated-insecure-static-crypto but you should move to a proper configuration using TLS as soon as possible.
Sep 15 11:11:34 fw-server openvpn(ns_testp2p)[21639]: Use --help for more information.
```
Tunnel Client logs :

```
Sep 15 11:11:31 fw-client openvpn(ns_ctestp2p)[1000]: DEPRECATED OPTION: The option --secret is deprecated.
Sep 15 11:11:31 fw-client openvpn(ns_ctestp2p)[1000]: Options error: DEPRECATION: No tls-client or tls-server option in configuration detected. OpenVPN 2.8
 will remove the functionality to run a VPN without TLS. See the examples section in the manual page for examples of a similar quick setup with peer-fingerprint. OpenVPN 2
[...]

PR #1901, opened 2026-09-16
Fix OpenVPN P2P tunnels not starting

gsanchietti:
Changes:

- patch upstream OpenVPN to support `allow-deprecated-insecure-static-crypto` option
- update all existing p2p tunnels
- set the option also inside new and migrated p2p tunnels

Related issue: #1899


Tbaile (approved):
I'd just remove the version bumps for the moment

PR #1904, opened 2026-09-17
feat(phonehome): return OpenVPN tunnel type

Tbaile:
Since the creation of https://github.com/NethServer/nethsecurity/issues/1899, we should be aware of how many tunnel types will be there in the future.


gsanchietti (approved):
Also verified.

Feel free to release

PR #1906, opened 2026-09-17
fix(openvpn): set default tun_mtu and mssfix values for migration

### LINKED ISSUE #1905: OpenVPN Road Warrior: missing MTU and MSS after migration import
When importing a NethServer 7 migration archive into NethSecurity 8, the OpenVPN Road Warrior configuration is restored without the `tun_mtu` and `mssfix` values.

As a consequence, the configuration cannot be saved from the UI until the missing values are set from the advanced options.

**Steps to reproduce**

* Configure an OpenVPN Road Warrior server on a NethServer 7 system
* Generate the migration archive
* Install a fresh NethSecurity 8 system and import the migration archive
* Open the migrated OpenVPN Road Warrior configuration
* Click `Save` without opening `Advanced options`

**Expected behavior**

The migrated OpenVPN Road Warrior configuration should contain the default `tun_mtu` and `mssfix` values and should be saved without requiring any additional action.

The resulting configuration should be consistent with an in-place migration.

**Actual behavior**

After importing the migration archive, `tun_mtu` and `mssfix` are missing from the OpenVPN Road Warrior configuration.

Clicking `Save` does not save the configuration and no validation error is visible.

After opening `Advanced options`, both fields are displayed as required and empty.

The problem does not occur with an in-place migration

**See also**

- https://mattermost.nethesis.it/nethesis/pl/sxdxhdux7jg6pdz18nq3kw97qo
- https://helpdesk.nethesis.it/a/tickets/205590

m-dilorenzi:
This PR contains the changes needed to fix the migration from NethSecurity 7.9 via archive upload, where the `tun_mtu` and `mssfix` values were not set correctly.

Changes:

- `ns-migration` migration scripts (`openvpn`, `openvpn_tunnels`) now set tun_mtu=1500 / mssfix=1450 when the NS7 export doesn't provide them
- `ns.ovpnrw` and `ns.ovpntunnel` apply the same defaults when reading an instance, covering already-migrated systems and restored backups.

Closes: https://github.com/NethServer/nethsecurity/issues/1905


Tbaile (approved):
You could use https://docs.python.org/3/builtins/stdtypes.html#dict.get default to remove some `if`s, otherwise LGTM.

Tbaile:
Pinging @gsanchietti on the review just to ask if he remembers something that might be breaking when migrating and setting these defaults. I have no clue what they were on NS7.

gsanchietti (approved):

